PR #11, opened 2015-04-25
Fix size calculation

jhiesey:
- fix type of ret in ostree_repo_get_commit_sizes()
 - fix size calculation to use gint64
 - fix number of files calculation summary command

[endlessm/eos-shell#4730]

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -491,10 +491,10 @@ ostree_repo_get_parent (OstreeRepo  *self)
   return self->parent_repo;
 }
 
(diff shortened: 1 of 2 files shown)


jhiesey:
I haven't yet verified that the data in the logs when updating is correct, but the summary command does work properly now.

dbnicholson:
Looks good, but could you break this up into 3 commits? I see 3 logical fixes: 64 bit for sizes, gsize instead of gboolean, and the calculation fix in the summary.

jhiesey:
Fixed

PR #12, opened 2015-05-08
Add commit size functionality

jhiesey:
This PR adds support for reading the upstream commit size format. This is the first step in implementing https://github.com/endlessm/eos-shell/issues/1965

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -309,14 +309,14 @@ commit_loose_object_trusted (OstreeRepo        *self,
 typedef struct
 {
   OstreeObjectType objtype;
-  gsize unpacked;
-  gsize archived;
+  guint64 unpacked;
+  guint64 archived;
 } OstreeContentSizeCacheEntry;
 
 static OstreeContentSizeCacheEntry *
 content_size_cache_entry_new (OstreeObjectType objtype,
-                              gsize            unpacked,
-                              gsize            archived)
+                              guint64          unpacked,
+                              guint64          archived)
 {
   OstreeContentSizeCacheEntry *entry = g_slice_new0 (OstreeContentSizeCacheEntry);
 
@@ -338,8 +338,8 @@ static void
 repo_store_size_entry (OstreeRepo       *self,
                        OstreeObjectType  objtype,
                        const gchar      *checksum,
-                       gsize             unpacked,
-                       gsize             archived)
+                       guint64           unpacked,
+                       guint64           archived)
 {
   if (G_UNLIKELY (self->object_sizes == NULL))
     self->object_sizes = g_hash_table_new_full (g_str_hash, g_str_equal,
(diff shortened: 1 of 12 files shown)


src/libostree/ostree-repo-commit.c
```diff
@@ -424,6 +429,136 @@ add_size_index_to_metadata (OstreeRepo        *self,
...
+  gs_unref_variant GVariant *metadata = NULL;
+  gs_unref_variant GVariant *sizes = NULL;
+
+  if (!ostree_repo_load_variant (self, OSTREE_OBJECT_TYPE_COMMIT, rev,
+                            &commit, error))
+    goto out;
+
+  metadata = g_variant_get_child_value (commit, 0);
+  if (!metadata)
+    goto out;
```
jhiesey: I realized that I'm forgetting to set `error` in a bunch of cases. I'll fix this next time I'm in the office.

dbnicholson:
Besides my comment on https://github.com/endlessm/ostree/commit/e6a8f2d3a996a9aec25d588423caa39ca989047c, I think the first 4 commits should go in immediately so we have a passing test suite again. I'm still looking at the rest implementing the metadata.

jhiesey:
@dbnicholson I created a separate PR (https://github.com/endlessm/ostree/pull/13) with just the test fixes so we can merge that first. I updated the commit message on the double-free bug there as well.

dbnicholson:
@jhiesey Great. I accepted #13. Please rebase this PR and let me know when you have it ready for review again.

src/libostree/ostree-repo-commit.c
```diff
@@ -406,6 +410,7 @@ add_size_index_to_metadata (OstreeRepo        *self,
           e_size = g_hash_table_lookup (self->object_sizes, e_checksum);
           _ostree_write_varuint64 (buffer, e_size->archived);
           _ostree_write_varuint64 (buffer, e_size->unpacked);
+          g_string_append_c(buffer, (guchar) e_size->objtype);
```
jhiesey: Come to think of it, do we really want to write the objtype directly into the sizes variant? It should be fine if the OstreeObjectType values are globally unique, but if e.g. values 5 and 6 are reused for something else upstream at some point we could be in trouble

dbnicholson: Indeed, that would be a problem. I've reworked the compat file stuff now so that it's all internal like the detached metadata and doesn't add core objects. Let's assume that we won't be adding objects to the ostree core for now.

BTW, space before paren for consistent code style.

src/libostree/ostree-repo-commit.c
```diff
@@ -308,16 +308,19 @@ commit_loose_object_trusted (OstreeRepo        *self,
 
 typedef struct
 {
-  gsize unpacked;
-  gsize archived;
+  OstreeObjectType objtype;
+  guint64 unpacked;
+  guint64 archived;
```
dbnicholson: You made these 64 bit, but all the entry functions are still using `gsize`. Was that intentional?

src/libostree/ostree-repo-commit.c
```diff
@@ -424,6 +429,139 @@ add_size_index_to_metadata (OstreeRepo        *self,
...
 static gboolean
+ostree_repo_commit_unpack_sizes (GVariant                    *entry,
+                                 OstreeContentSizeCacheEntry *sizes,
+                                 char                        *csum)
+{
+  gboolean ret = FALSE;
+  const guchar *buffer;
+  gsize bytes_read = 0;
+
+  gsize object_size = g_variant_get_size(entry);
```
dbnicholson: Space before paren. I also think it's more typical to keep all the declarations in the block at the top.

src/libostree/ostree-repo-commit.c
```diff
@@ -424,6 +429,139 @@ add_size_index_to_metadata (OstreeRepo        *self,
...
+
+  if (!_ostree_read_varuint64 (buffer, object_size, &(sizes->unpacked), &bytes_read))
+    goto out;
+  buffer += bytes_read;
+  object_size -= bytes_read;
+
+  if (object_size < 1)
+    goto out;
+
+  if (*buffer < 1 || *buffer > 4)
```
dbnicholson: This should probably use the `OSTREE_OBJECT_TYPE` macros.

src/libostree/ostree-repo-commit.c
```diff
@@ -424,6 +429,139 @@ add_size_index_to_metadata (OstreeRepo        *self,
...
+              if (entry.objtype == OSTREE_OBJECT_TYPE_FILE)
+                n_files++;
+            }
+        }
+      else
+        {
+          if (error && *error)
+            g_debug ("has_object test failed: %s", (*error)->message);
+          goto out;
+        }
```
dbnicholson: Is this block needed? Typical ostree coding would have this just `goto out` and let the error bubble up. You could prefix it if desired. I think the more natural ostree code would be like:
```
if (!ostree_repo_has_object(...))
  goto out;
if (!exists)
  add stuff;
```

src/libostree/ostree-repo-commit.c
```diff
@@ -424,6 +429,139 @@ add_size_index_to_metadata (OstreeRepo        *self,
...
+      gboolean exists;
+
+      if (!ostree_repo_commit_unpack_sizes (object, &entry, csum))
+      {
+        g_variant_unref (object);
+        g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                     "Invalid object size metadata");
+        goto out;
+      }
+      g_variant_unref (object);
```
dbnicholson: It looks like object is always unrefed. Can it just be marked with `gs_unref_variant` in the declaration?

src/libostree/ostree-repo-commit.c
```diff
@@ -424,6 +429,139 @@ add_size_index_to_metadata (OstreeRepo        *self,
...
+  guint64 t_unpacked = 0;
+  gsize t_files = 0;
+  GVariantIter obj_iter;
+  GVariant *object;
+  gs_unref_variant GVariant *commit = NULL;
+  gs_unref_variant GVariant *metadata = NULL;
+  gs_unref_variant GVariant *sizes = NULL;
+
+  if (!ostree_repo_load_variant (self, OSTREE_OBJECT_TYPE_COMMIT, rev,
+                            &commit, error))
```
dbnicholson: Indentation off here.

src/libostree/ostree-repo-commit.c
```diff
@@ -655,9 +792,6 @@ write_object (OstreeRepo         *self,
           gs_unref_object GConverter *zlib_compressor = NULL;
           gs_unref_object GOutputStream *compressed_out_stream = NULL;
 
-          if (self->generate_sizes)
-            indexable = TRUE;
```
dbnicholson: This change isn't really explained in any of the commit messages. It seems to me there are 2 implications of it:

1. Sizes will be generated for not only file objects, but also metadata objects.
2. Sizes will be generated for all repository types, not just archive-z2.

Is that correct? Are there any other implications of this change?

src/ostree/ot-builtin-size-summary.c
```diff
@@ -0,0 +1,138 @@
...
+  gs_unref_variant GVariant *index = NULL;
+  gs_free gchar *refspec = NULL;
+  gs_free gchar *revision = NULL;
+  gint64 archived = 0;
+  gint64 unpacked = 0;
+  gsize  fetch_needed = 0;
+  gint64 new_archived = 0;
+  gint64 new_unpacked = 0;
+
+  context = g_option_context_new("[REMOTE] BRANCH - Display the summary for branch");
```
dbnicholson: Space before paren.

src/ostree/ot-builtin-size-summary.c
```diff
@@ -0,0 +1,138 @@
...
+      goto out;
+    }
+
+  if (ostree_repo_get_commit_sizes (repo, revision,
+                                          &new_archived,
+                                          &new_unpacked,
+                                          &fetch_needed,
+                                          &archived, &unpacked,
+                                          &entries,
+                                          cancellable, error))
```
dbnicholson: Indentation wrong here.

src/ostree/ot-builtin-size-summary.c
```diff
@@ -0,0 +1,138 @@
...
+               refspec,
+               entries - fetch_needed, entries,
+               archived - new_archived, archived,
+               unpacked - new_unpacked, unpacked);
+    }
+  else if (!*error) // No error found, 0 entries => no index available
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
+                   "Refspec '%s' has no index, cannot summarise", refspec);
+      goto out;
```
dbnicholson: I'm not sure why the error is being overriden here. If `ostree_repo_get_commit_sizes` returned FALSE, error should contain the actual error already.

src/ostree/ot-builtin-size-summary.c
```diff
@@ -0,0 +1,138 @@
...
+      if (!ostree_repo_pull (repo, remote, pullrefs, flags, NULL, cancellable, error))
+        goto out;
+
+      if (!ostree_repo_resolve_rev (repo, refspec, FALSE, &revision, error))
+        goto out;
+    }
+
+  if (!revision)
+    {
+      if (!*error)
```
dbnicholson: I don't think this needs to be checked. `error` should only be set if the previous calls to `ostree_repo_resolve_rev` failed, in which case you'll never get here.

dbnicholson:
@jhiesey Looks good in general, but I added some comments to address. I realize some of this code wasn't yours, but considering we want to upstream this, we should try to make it clean.

dbnicholson:
Before this gets merged, we should have a look at the code on https://bugzilla.gnome.org/show_bug.cgi?id=709050 and see if we can keep the metadata compatible.

src/libostree/ostree-repo-commit.c
jhiesey: I still need `g_variant_unref` in the loop though, right?

src/libostree/ostree-repo-commit.c
jhiesey: Not changing these was definitely sloppy, but unfortunately due to the return type of `g_output_stream_splice` we still can't handle files with a compressed size that doesn't fit in a signed 32 bit integer. I'll change everything that I can, which will make at least the uncompressed sizes always right, but what can we do about that other case?

src/ostree/ot-builtin-size-summary.c
jhiesey: Whoops, that was left over from when `ostree_repo_get_commit_sizes` returned something different.

src/libostree/ostree-repo-commit.c
dbnicholson: You're right. In my previous reading, I didn't realize that object was being assigned in the loop iterator. You do indeed need to unref it in the loop somewhere, and I can't see an easy way not to duplicate the unref in the error case.

jhiesey:
The metadata should be compatible with upstream. The only interesting issue is the one raised here: https://bugzilla.gnome.org/show_bug.cgi?id=709050#c15

I think Colin might be right to implement the commit only fetch in a different way. I could try starting with that patch instead of my implementation and make sure it works.

dbnicholson:
Looks good, just a couple more comments. After that, I'd really love to see this attached to the upstream bug to see if there are any objections upstream. We can probably merge here in the meantime, but I'd really want to have upstream review before merging to master.

dbnicholson:
CR pass!

PR #14, opened 2015-05-14
tests: Link test-gpg-verify-result with gpgme

dbnicholson:
This test uses gpgme directly to verify the signatures, so it needs to
find the gpgme headers and link with gpgme to ensure the linker can
resolve the symbols.


dbnicholson:
@jhiesey I needed this to run `make check`. Can you have a look?

jhiesey:
Looks good.

PR #15, opened 2015-05-15
Maintain compat files

jhiesey:
I'm creating this PR so I can comment on @dbnicholson's changes


src/libostree/ostree-core.h
```diff
@@ -73,14 +77,14 @@ typedef enum {
  *
  * Returns: %TRUE if object type is metadata
  */
-#define OSTREE_OBJECT_TYPE_IS_META(t) (t >= 2 && t <= 4)
+#define OSTREE_OBJECT_TYPE_IS_META(t) (t >= 2 && t <= 6)
```
jhiesey: Can't we use the enum constants here?

dbnicholson:
So, after thinking on this a bit more, I think it would be best to not add these as core ostree objects. It makes the code easier to manage because I don't have to add a bunch of hacks to the core object code for optional files. It also means we don't add core object types that would be incompatible with upstream if they added any object types later.

I opened  #16 now and propose that we close this one.

dbnicholson:
Since #16 got merged, I'm going to go ahead and close this request.

PR #16, opened 2015-05-21
Handle compat files

dbnicholson:
Read, write and delete the optional compat objects like the detached metadata. In addition, create the compat signature files when committing. Still to do - add tests and verify commits if they only have the old style signature.

[endlessm/eos-shell#5083]


dbnicholson:
If this is acceptable, I think #15 should be closed and the https://github.com/endlessm/ostree/tree/compat-files-read and https://github.com/endlessm/ostree/tree/compat-signature-write branches should be deleted. All the work from those is included in a compatible format here.

src/libostree/ostree-repo-commit.c
```diff
@@ -2017,6 +2017,207 @@ ostree_repo_write_commit_detached_metadata (OstreeRepo      *self,
...
+    }
+  else
+    {
+      if (!g_file_replace_contents (signature_path,
+                                    signature_buf,
+                                    signature_size,
+                                    NULL, FALSE, 0, NULL,
+                                    cancellable, error))
+        {
+          g_prefix_error (error, "Failed to write compat signature: ");
```
plward11: @dbnicholson Do you need to clear the error here?

dbnicholson: I don't think so. The error is prefixed with another message, but it's the same error set in the failing call. This is pretty topical in ostree code.

PR #26, opened 2015-08-04
Revert "Revert "Don't make the root immutable""

dbnicholson:
This reverts commit be1131def9a53804757866b448b6bebc30de7000.
Unfortunately, this breaks systemd's readahead implementation, which
tries to create /.readahead and /.readahead.new.

[endlessm/eos-shell#5190]


sgnn7:
@dbnicholson CR pass.

PR #31, opened 2015-09-24
Revert "Revert "Revert "Don't make the root immutable"""

jhiesey:
This reverts commit 3c3a52793c09da29c02f36d67a953ed234a8726c.
Now that /.readahead has been moved to /var/.readahead, systemd's
readahead implementation should not be broken by an immutable
root.

[endlessm/eos-shell#5190]


jhiesey:
This should be merged after https://github.com/endlessm/systemd/pull/17

PR #32, opened 2015-10-02
static-delta: Set error on bsdiff failure

jhiesey:


jhiesey:
This is incorporated into the fixes in #34

PR #35, opened 2015-10-21
static-delta: Fix annotation on ostree_repo_list_static_delta_names

jhiesey:
out_deltas should be (transfer-container)

[endlessm/eos-shell#5572]


cosimoc:
Merging this, as it was accepted upstream already.

PR #36, opened 2015-10-27
sysroot: Write symlinks before calling fsync(), then rename after

cosimoc:
There might be a race here in that we create new symlink files *after*
calling `syncfs`, and they are not guaranteed to end up on disk.

Rework the code so that we create symlinks before, and then only
rename them after (and `fsync()` the directory for good measure).

Additional-fixes-by: Giuseppe Scrivano <gscrivan@redhat.com>
Tested-by: Giuseppe Scrivano <gscrivan@redhat.com>

This still needs verification that we're fixing a real bug; but I'm
fairly confident this won't make the fsync situation worse.

https://bugzilla.gnome.org/show_bug.cgi?id=755595

[endlessm/eos-shell#5756]


src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1073,40 +1124,79 @@ swap_bootlinks (OstreeSysroot *self,
...
+      if (!_ostree_sysroot_read_current_subbootversion (self, bootversion, &old_subbootversion,
+                                                        cancellable, error))
+        goto out;
+    }
+  else
+    old_subbootversion = self->subbootversion;
+
+  new_subbootversion = old_subbootversion == 0 ? 1 : 0;
+
+  ostree_subbootdir_name = g_strdup_printf ("boot.%d.%d", bootversion, new_subbootversion);
```
jhiesey: I wish the previous 17 lines weren't duplicated from `create_new_bootlinks` above

jhiesey:
It looks ok to me as long as later writes to one fs can't succeed when earlier ones fail (which I would expect journaling to guarantee).

It's also easier to reason about than the previous design.

PR #41, opened 2016-06-21
Rebase t11978

smspillaz:
Rebase on OSTree 2016.6

This includes the proposed fix for https://phabricator.endlessm.com/T11978, being 535033a4f01f772e696ad2201b34403203b98bea

PR #42, opened 2016-06-27
Rebase t11978 dbn

smspillaz:
This is a continuation of #41

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -252,7 +252,7 @@ ALL_LOCAL_RULES += tests/libreaddir-rand.so
 CLEANFILES += tests/libreaddir-rand.so tests/ostree-symlink-stamp tests/ostree
 
 tests/ostree-symlink-stamp: Makefile
-	@real_bin=`cd $(top_builddir) && libtool --mode=execute echo ostree`; \
+	@real_bin=`cd $(top_builddir) && ./libtool --mode=execute echo ostree`; \
 	ln -sf "$${real_bin}" tests/ostree; \
 	touch $@
 
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -252,7 +252,8 @@ ALL_LOCAL_RULES += tests/libreaddir-rand.so
 CLEANFILES += tests/libreaddir-rand.so tests/ostree-symlink-stamp tests/ostree
 
 tests/ostree-symlink-stamp: Makefile
-	@real_bin=`cd $(top_builddir) && ./libtool --mode=execute echo ostree`; \
+	@set -e; \
+	real_bin=`cd $(top_builddir) && ./libtool --mode=execute echo ostree`; \
 	ln -sf "$${real_bin}" tests/ostree; \
 	touch $@
 
--- a/src/libostree/libostree.sym
+++ b/src/libostree/libostree.sym
@@ -352,6 +352,6 @@ global:
 /*
 LIBOSTREE_2016.7 {
 global:
-        ostree_some_new_symbol;
+        someostree_some_new_symbol;
 } LIBOSTREE_2016.6;
 */
--- a/tests/test-admin-deploy-karg.sh
+++ b/tests/test-admin-deploy-karg.sh
@@ -43,10 +43,9 @@ assert_file_has_content sysroot/boot/loader/entries/ostree-testos-0.conf 'option
 echo "ok deploy with --karg, but same config"
 
 ${CMD_PREFIX} ostree admin deploy --karg-proc-cmdline --os=testos testos:testos/buildmaster/x86_64-runtime
-# Here we're asserting that the *host* system has a root= kernel
-# argument.  I think it's unlikely that anyone doesn't have one, but
-# if this is not true for you, please file a bug!
-assert_file_has_content sysroot/boot/loader/entries/ostree-testos-0.conf 'options.*root=.'
+for arg in $(cat /proc/cmdline); do
+    assert_file_has_content sysroot/boot/loader/entries/ostree-testos-0.conf "options.*$arg"
+done
 
 echo "ok deploy --karg-proc-cmdline"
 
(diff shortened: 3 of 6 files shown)


src/libostree/ostree-repo.c
```diff
@@ -5036,3 +5036,50 @@ _ostree_repo_allocate_tmpdir (int tmpdir_dfd,
...
+
+      _ostree_loose_path_with_extension (compat_sig_loose, commit_checksum,
+                                         "sig");
+      do
+        res = unlinkat (self->objects_dir_fd, compat_sig_loose, 0);
+      while (G_UNLIKELY (res == -1 && errno == EINTR));
+      if (G_UNLIKELY (res == -1))
+        {
+          gs_set_error_from_errno (error, errno);
+          goto out;
```
smspillaz: Probably not important for this rebase, but something I noted here is that there's a lot of duplicated code between here and b460128e17510b8c0afc425f880490eca227b8aa, maybe there's an opportunity to create a new function out of it.

src/libostree/ostree-repo-commit.c
```diff
@@ -424,6 +424,134 @@ add_size_index_to_metadata (OstreeRepo        *self,
...
+  if (!_ostree_read_varuint64 (buffer, object_size, &(sizes->unpacked), &bytes_read))
+    goto out;
+  buffer += bytes_read;
+  object_size -= bytes_read;
+
+  if (object_size < 1)
+    goto out;
+
+  if (*buffer < OSTREE_OBJECT_TYPE_FILE || *buffer > OSTREE_OBJECT_TYPE_LAST)
+    goto out;
```
smspillaz: Not important for this rebase, but these would be better implemented using early returns as opposed to `goto`

src/libostree/ostree-repo-commit.c
```diff
@@ -424,6 +424,134 @@ add_size_index_to_metadata (OstreeRepo        *self,
...
+
+      t_archived += entry.archived;
+      t_unpacked += entry.unpacked;
+
+      if (entry.objtype == OSTREE_OBJECT_TYPE_FILE)
+        t_files++;
+
+      if (!ostree_repo_has_object (self, entry.objtype,
+                                   csum, &exists, cancellable, error))
+        goto out;
```
smspillaz: Same as earlier, use early returns instead of `goto` here.

src/libostree/ostree-repo.c
dbnicholson: Could be. They're a little different in that this one is looking for a particular signature while the other one is deleting the whole signature file regardless of what's in it since the commit is being deleted. I guess you could make a common subroutine that does the actual deletion, though. I think I was following the existing style.

src/libostree/ostree-repo-commit.c
```diff
@@ -770,10 +895,10 @@ write_object (OstreeRepo         *self,
               /* Don't close the base; we'll do that later */
               g_filter_output_stream_set_close_base_stream ((GFilterOutputStream*)compressed_out_stream, FALSE);
               
-              unpacked_size = g_output_stream_splice (compressed_out_stream, file_input,
-                                                      0, cancellable, error);
-              if (unpacked_size < 0)
+              if (g_output_stream_splice (compressed_out_stream, file_input, 0, cancellable, error) < 0)
```
smspillaz: If we go with upstream's version here we can remove the extract `g_file_info_get_size`

src/libostree/ostree-repo-commit.c
dbnicholson: I think I (or John) was trying to stick with ostree style where there's very rarely early returns. This could _could_ go upstream (https://bugzilla.gnome.org/show_bug.cgi?id=756266), so I'd defer to upstream style.

src/libostree/ostree-repo-commit.c
```diff
@@ -819,7 +945,7 @@ write_object (OstreeRepo         *self,
 
   g_assert (actual_checksum != NULL); /* Pacify static analysis */
           
-  if (indexable && temp_file_is_regular)
```
smspillaz: Might want to use upstream's variable here (`indexable` instead of `self->generate_sizes`)

src/libostree/ostree-repo-commit.c
dbnicholson: Again, this is just following upstream's convention.

src/libostree/ostree-repo-commit.c
smspillaz: I figured that was the case. It'd be good if upstream could eventually move away from this style of error handling, but upstream style trumps.

src/libostree/ostree-repo-commit.c
dbnicholson: I actually think this is intentional, IIRC. `g_output_stream_splice` returns a `gssize`, which will be 32 bit on i386. That limits the size of the unpacked size to 4 GB. On the other hand, `g_file_info_get_size` returns a `goffset`, which is always 64 bit. I remember @jhiesey talking to me about this issue with GOutputStream.

smspillaz:
LGTM otherwise

src/libostree/ostree-repo-commit.c
dbnicholson: I _think_ this is intentional. Before this change, `indexable` was only set for `objtype == OSTREE_OBJECT_TYPE_FILE` and `repo_mode == OSTREE_REPO_MODE_ARCHIVE_Z2`. However, what @jhiesey was trying to achieve was getting the size for _all_ files, including metadata. That would happen in the `else` from `objtype == OSTREE_OBJECT_TYPE_FILE`.

src/libostree/ostree-repo-commit.c
dbnicholson: I'm pretty sure this is a core tenet of Colin's style. And it does have one a big advantage over early returns: the return path is always the same. If you add a new cleanup path, you don't have to go back and analyze all the early returns to see how they'd need to handle it.

dbnicholson:
I think I responded to all the comments. Let me know if there's anything that you definitely want fixed before pushing.

dbnicholson:
Pushed this to master. Closing.

PR #44, opened 2016-07-31
Don't restrict the size of the legacy .sizes2 file

dbnicholson:
The legacy object sizes format is not very space efficient,
causing it to easily exceed OSTREE_MAX_METADATA_SIZE.

To avoid bailing out, don't verify the size of that file.

This got missed on the last rebase. I hit this on the server when pulling old commits. @smspillaz could you review?

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1435,7 +1435,7 @@ enqueue_one_object_request (OtPullData        *pull_data,
   expected_max_size_p = (fetchtype != OSTREE_FETCH_OBJECT_CORE) ? NULL : g_hash_table_lookup (pull_data->expected_commit_sizes, checksum);
   if (expected_max_size_p)
     expected_max_size = *expected_max_size_p;
-  else if (is_meta)
+  else if (is_meta && fetchtype != OSTREE_FETCH_OBJECT_COMPAT_SIZES)
     expected_max_size = OSTREE_MAX_METADATA_SIZE;
   else
     expected_max_size = 0;


smspillaz:
Accepting, since this was missed in the last rebase, however, there is something I don't understand and I have commented on the diff about that.

src/libostree/ostree-repo-pull.c
```diff
@@ -1435,7 +1435,7 @@ enqueue_one_object_request (OtPullData        *pull_data,
   expected_max_size_p = (fetchtype != OSTREE_FETCH_OBJECT_CORE) ? NULL : g_hash_table_lookup (pull_data->expected_commit_sizes, checksum);
   if (expected_max_size_p)
     expected_max_size = *expected_max_size_p;
-  else if (is_meta)
+  else if (is_meta && fetchtype != OSTREE_FETCH_OBJECT_COMPAT_SIZES)
```
smspillaz: The PR message says that the format could exceed `OSTREE_MAX_METADATA_SIZE`, but here it looks like we’re assigning `OSTREE_MAX_METADATA_SIZE` to `expected_max_size` for `OSTREE_FETCH_COMPAT_OBJECT_SIZES`? I'm not sure I understand what's going on here. I trust that this works since it was in a prior version, but it might be enlightening for me if I knew why the assignment fixed the issue.

dbnicholson: It gets assigned to `OSTREE_MAX_METADATA_SIZE` if `fetchtype != OSTREE_FETCH_COMPAT_OBJECT_SIZES`. Otherwise, it falls through to `0`, which gets interpreted as unlimited size.

PR #46, opened 2016-08-31
repo-pull: properly store the cancellable

GeorgesStavracas:
OSTree function ostree_repo_pull_with_options starts a
series of operations that makes heavy use of the PullData's
cancellable.

This isn't effective, however, since nowhere in the code
the OtPullData.cancellable field is set. This is visible,
for example, when trying to cancel a Flatpak pull and nothing
happens, because the cancellable is not properly passed
to the pull data.

Fix that by setting the cancellable field of the pull data. It
owns a reference for safety reasons, and unreferences it at the
end of the operation.

https://phabricator.endlessm.com/T12885


dbnicholson:
Looks good, nice find. Corresponds to upstream's https://github.com/ostreedev/ostree/commit/cafc517c6bb93bb45777b6969125cb9599f217f6.

PR #56, opened 2016-12-09
Various fixes

dbnicholson:
A few fixes I found while working some ostree scripts for the image builder. The first commit got in upstream, and I'd like it here so I can actually read the remote summary data from python. The next 2 fix our non-upstreamed size summary to actually work correctly when the sizes information is not there. The last annotates the function so I can get commit sizes out from python.

### DIFF
--- a/src/ostree/ot-builtin-size-summary.c
+++ b/src/ostree/ot-builtin-size-summary.c
@@ -30,6 +30,7 @@
 gboolean
 ostree_builtin_size_summary (int argc, char **argv, GCancellable *cancellable, GError **error)
 {
+  gboolean ret = FALSE;
   GOptionContext *context;
   OstreeRepo *repo;
   gsize entries = 0;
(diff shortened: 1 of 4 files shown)


dbnicholson:
@mariospr perhaps you can take a look at these changes? I can open an issue if you prefer, but these were a bit random.

cosimoc:
@mariospr Sorry! I did not see the other comment by @dbnicholson and merged this... Feel free to also take a look and propose fixes if needed

mariospr:
@cosimoc No problem, my only comment was more kind of a question for @dbnicholson and not going to be blocking in any case. I seriously don't think it will make a difference or be more correct either way.

src/ostree/ot-builtin-size-summary.c
```diff
@@ -98,26 +99,27 @@ ostree_builtin_size_summary (int argc, char **argv, GCancellable *cancellable, G
...
+  g_print ("Summary for refspec %s:\n"
+           "  files: %" G_GSIZE_FORMAT "/%" G_GSIZE_FORMAT " entries\n"
+           "  archived: %" G_GINT64_FORMAT "/%" G_GINT64_FORMAT "\n"
+           "  unpacked: %" G_GINT64_FORMAT "/%" G_GINT64_FORMAT "\n",
+           refspec,
+           entries - fetch_needed, entries,
+           archived - new_archived, archived,
+           unpacked - new_unpacked, unpacked);
 
+  ret = TRUE;
```
dbnicholson: Sorry, I should have commented this a bit more. If the sizes information is available, `entries` is the number of files from the commit that are present in the repo. Since you might have a commit without files or you might have only the commit and none of the objects, this isn't really an error. The real error is if the sizes data is not available at all in the commit metadata (at least the way the code is now).

Does that make sense?

mariospr: Yes, makes sense. Thanks for the clarification, as that was not obvious by reading the code change, which is not technically not equivalent to what it was before.

PR #57, opened 2016-12-14
static-delta: Pretend that world unreadable objects are new objects

mariospr:
This will prevent including in the delta the bits to update files that
are not world readable, so that we don't run into a permissions problem
when applying the deltas from a bare-user repository that has a bare
repository set as its parent.

This is the case when updating flatpak apps/runtimes, as the temporary
directory created in ~/.local/share/flatpak/system-cache will be of
type bare-user with its parent set to /var/lib/flatpak which is a bare
repository in our case, as it's shared with the one at /ostree/repo.

https://phabricator.endlessm.com/T14159

### DIFF
--- a/src/libostree/ostree-repo-static-delta-compilation.c
+++ b/src/libostree/ostree-repo-static-delta-compilation.c
@@ -865,6 +865,25 @@ process_one_bsdiff (OstreeRepo                       *repo,
   return ret;
 }
 
+static gboolean
+check_object_world_readable (OstreeRepo   *repo,
+                             const char   *checksum,
+                             gboolean     *out_readable,
+                             GCancellable *cancellable,
+                             GError      **error)
+{
+  g_autoptr(GFileInfo) finfo = NULL;
+  guint32 mode;
+
+  if (!ostree_repo_load_file (repo, checksum, NULL, &finfo, NULL,
+                              cancellable, error))
+    return FALSE;
+
+  mode = g_file_info_get_attribute_uint32 (finfo, "unix::mode");
+  *out_readable = (mode & S_IROTH);
+  return TRUE;
+}
+
 static gboolean 
 generate_delta_lowlatency (OstreeRepo                       *repo,
                            const char                       *from,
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-static-delta-compilation.c
```diff
@@ -865,6 +865,24 @@ process_one_bsdiff (OstreeRepo                       *repo,
...
+static gboolean
+check_object_world_readable (OstreeRepo   *repo,
+                             const char   *checksum,
+                             gboolean     *out_readable,
+                             GCancellable *cancellable,
+                             GError      **error)
+{
+  g_autoptr(GFileInfo) finfo = NULL;
+  guint32 mode;
+  if (!ostree_repo_load_file (repo, checksum, NULL, &finfo, NULL,
```
dbnicholson: Blank line between declarations and code is typical.

src/libostree/ostree-repo-static-delta-compilation.c
```diff
@@ -997,6 +1015,18 @@ generate_delta_lowlatency (OstreeRepo                       *repo,
       const char *from_checksum = value;
       ContentRollsum *rollsum;
       ContentBsdiff *bsdiff;
+      gboolean from_world_readable = FALSE;
+      gboolean to_world_readable = FALSE;
+
+      if (!check_object_world_readable (repo, from_checksum, &from_world_readable, cancellable, error))
```
dbnicholson: I think a comment here explaining why non-world-readable files are being skipped would be good.

src/libostree/ostree-repo-static-delta-compilation.c
```diff
@@ -1097,8 +1128,12 @@ generate_delta_lowlatency (OstreeRepo                       *repo,
                                            NULL, &uncompressed_size,
                                            cancellable, error))
         goto out;
-      if (builder->min_fallback_size_bytes > 0 &&
-          uncompressed_size > builder->min_fallback_size_bytes)
+
+      if (!check_object_world_readable (repo, checksum, &world_readable, cancellable, error))
```
dbnicholson: Comment here would be good, too.

src/libostree/ostree-repo-static-delta-compilation.c
```diff
@@ -1097,8 +1128,12 @@ generate_delta_lowlatency (OstreeRepo                       *repo,
...
+        goto out;
+
+      if (!world_readable || (builder->min_fallback_size_bytes > 0 &&
```
dbnicholson: I think aligning the second condition on the next liine is more typical here:
```
if (!world_readable ||
    (other_cond &&
     another_cond))
  {
```

src/libostree/ostree-repo-static-delta-compilation.c
mariospr: Agreed, just slipped in.

src/libostree/ostree-repo-static-delta-compilation.c
mariospr: Seems sane, I'll add it.

src/libostree/ostree-repo-static-delta-compilation.c
mariospr: Ok.

src/libostree/ostree-repo-static-delta-compilation.c
mariospr: My WebKit past is haunting me again! Will change.

dbnicholson:
Can you change this to a straight cherry pick of upstream's [722e143](https://github.com/ostreedev/ostree/commit/722e143ee1109717064dc7ef06777f66a80116dd)? Then we include the info at the bottom of the commit message.

mariospr:
No problem, although I don't think we need to include anything because apparently I got the link to phabricator already in the commit upstream by mistake.

PR #59, opened 2017-02-24
fetcher: Avoid races inside libsoup input streams

krnowak:
When fetching data to membuf we were moving the input stream we got
from the soup session to the other thread where the stream was closed
after splicing it into the memory output stream. Current libsoup from
master does not quite handle it gracefully as there are some races
inside which may lead to the hang. The bug for libsoup was reported,
but the fix is still not agreed upon.

https://bugzilla.gnome.org/show_bug.cgi?id=778135

To work it around we do the splicing in the same thread where the
download happened, so the main thread just receives the memory buffer.

https://phabricator.endlessm.com/T15146

### DIFF
--- a/src/libostree/ostree-fetcher.c
+++ b/src/libostree/ostree-fetcher.c
@@ -1022,6 +1022,42 @@ on_stream_read (GObject        *object,
 }
 
 static void
+request_body_spliced (GObject      *object,
+                      GAsyncResult *result,
+                      gpointer      user_data)
+{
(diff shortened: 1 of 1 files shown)


cgwalters:
Why not file this at https://github.com/ostreedev/ostree/pulls ?

krnowak:
@cgwalters: Endless ostree is based on some older version of the upstream ostree. In upstream, libsoup is already used in only one thread.

PR #60, opened 2017-03-07
libostree: Allow compression level to be set for archive-z2 stream

pwithnall:
Add a ostree_raw_file_to_archive_z2_stream_with_options() variant of
ostree_raw_file_to_archive_z2_stream(), to allow a compression-level
option to be passed in and passed through to zlib.

This is useful when building archive-z2 files on the fly for
transmission over a non-bandwidth-limited channel, such as a local
network. In this case, CPU time is more valuable than bandwidth, so we
want a low compression level.

(Non-trivial cherry-pick.)

Closes: #721
Approved by: cgwalters

https://phabricator.endlessm.com/T15909


src/libostree/libostree.sym
```diff
@@ -373,12 +373,10 @@ global:
  *                         NOTE NOTE NOTE
  */
 
-/* Remove comment when first new symbol is added
-LIBOSTREE_2016.XX {
+LIBOSTREE_2017.3 {
```
dbnicholson: Uh, I don't think we want the symbol version to be 2017.3 on our version, do we? Our version is 2016.15.

pwithnall: I was a little hesitant about this, but it seems to work. My thinking regarding leaving it as `2017.3` rather than adding it in a `2016.15` section is that it’s the same as upstream — so when we upgrade OSTree, the symbol definition won’t magically switch from the latter to the former.

I admit I have not had a chance to re-read about this style of symbol versioning. I will do that before considering this task closed.

dbnicholson: Right. It definitely makes upgrading easier. One thing about the symbol versions is that they also get used in the debian symbols file - https://github.com/endlessm/ostree/blob/debian-master/debian/libostree-1-1.symbols. Since that wasn't updated, nothing will happen (except for [dpkg-gensymbols complaining about it](https://obs-master.endlessm-sf.com/package/live_build_log?arch=x86_64&package=ostree&project=eos%3Amaster%3Aendless&repository=endless)), but in theory it would be nice if it was so that programs using libostree get a proper versioned dependency if they use this new symbol.

Meh. I guess it's fine the way it is.

pwithnall: Ah, indeed, I keep forgetting about the Debian stuff when it’s in a separate `debian-master` branch. My bad.

Since the customisation is within our Debian package of OSTree, we could add the new symbol to the Debian symbols file without causing upgrade problems in future. I’ll file a follow-up PR to do this when I get a chance.

pwithnall: `debian-master` change here: https://github.com/endlessm/ostree/pull/62

PR #63, opened 2017-05-10
Fixes for pull service

dbnicholson:
A few fixes backported from upstream to be used in the new ostree-receive script. These will eventually go into the server package, but review is more convenient on master.

https://phabricator.endlessm.com/T5939


mariospr:
The three commits look good to me. Merging

PR #64, opened 2017-05-16
Use dh_auto_test for testing

krnowak:
It should be clever enough to call make check, but it also respects
the DEB_BUILD_OPTIONS environment variable, so I can skip the tests
when doing scratch builds on my computer.

### DIFF
--- a/debian/test.sh
+++ b/debian/test.sh
@@ -7,7 +7,7 @@ export VERBOSE=1
 try_tests=5
 
 failed=0
-make check || failed=1
+dh_auto_test || failed=1
 
 if [ "$failed" -gt 0 ]; then
     [ "$failed" -eq 0 ] || echo "Test failed! Checking how reproducible it is..."


dbnicholson:
Isn't this handled by `override_dh_auto_test` in `debian/rules`?

dbnicholson:
Or maybe that override needs to handle the option manually. I think that would be the more appropriate place to make any changes.

krnowak:
Alright, will update it tomorrow.

dbnicholson:
Bonus points if you file upstream in Debian. We follow the upstream packaging pretty closely and Simon McVittie is pretty responsive.

krnowak:
Alright, filed - https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=862803

dbnicholson:
Thanks, that looks good except for `s/ifeq/ifneq/` (comment should be coming to the Debian BTS soon). We can apply that locally if you want, too, but I'd like to wait a day to see if Simon responds.

krnowak:
I actually think `ifeq` is correct. :P The reasoning should be on the BTS shortly.

Anyway, no need to hurry with applying that locally - I just patched that thing myself and I'm fine.

dbnicholson:
Oh, right! You're using the conditional to say "yes, run the tests". This should only be done if `nocheck` is _not_ in `DEB_BUILD_OPTIONS`.

krnowak:
I'm closing it - the (different) fix I have made for upstream was just recently ported to debian-master branch.

PR #68, opened 2017-05-24
T16909

GeorgesStavracas:
Backport of some upstream patches that land the new atomic API. These patches should be dropped on a further rebase.

https://phabricator.endlessm.com/T16909


src/libostree/ostree-async-progress.c
```diff
@@ -204,26 +205,45 @@ ostree_async_progress_get_status (OstreeAsyncProgress       *self)
...
   g_mutex_lock (&self->lock);
 
   if (self->dead)
     goto out;
 
-  if (g_hash_table_lookup_extended (hash, qkey, NULL, &orig_value))
+  if (g_hash_table_lookup_extended (self->values, qkey, NULL, (gpointer *) &orig_value))
     {
-      if (orig_value == value)
+      if (g_variant_equal (orig_value, new_value))
```
joaquimrocha: These could have been condensed in one _if_ using _&&_.

src/libostree/ostree-async-progress.c
```diff
@@ -157,6 +157,68 @@ ostree_async_progress_get_uint64 (OstreeAsyncProgress       *self,
...
+{
+  va_list ap;
+  const char *key, *format_string;
+
+  g_mutex_lock (&self->lock);
+  va_start (ap, self);
+
+  for (key = va_arg (ap, const char *), format_string = va_arg (ap, const char *);
+       key != NULL;
+       key = va_arg (ap, const char *), format_string = va_arg (ap, const char *))
```
joaquimrocha: IMO this would be more readable if this was:
```
while ((key = va_arg (ap, const char *)) != NULL) {
    format_string = va_arg (ap, const char *);
    ...
}
```

PR #69, opened 2017-05-31
Don't remove ostree-boot files

GeorgesStavracas:
These files now belong to the ostree-boot package but,
since it'd require considerable effort in updating core
packages.

https://phabricator.endlessm.com/T17204

### DIFF
--- a/debian/ostree.install
+++ b/debian/ostree.install
@@ -1,5 +1,11 @@
+etc/dracut.conf.d/ostree.conf
 etc/ostree
+lib/systemd/system/ostree*.service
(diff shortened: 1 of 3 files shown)


debian/ostree.install
```diff
@@ -1,5 +1,10 @@
+etc/dracut.conf.d/ostree.conf
 etc/ostree
+lib/systemd/system/ostree*.service
```
dbnicholson: It looks ostree is now using a systemd generator to handle mounting of `/var` during boot, so `lib/systemd/system-generators/ostree-system-generator` will also be needed here.

PR #72, opened 2017-06-05
Move ostree_async_progress_* symbols to 2017.3

dbnicholson:
Unfortunately, when we backported these functions, we put them in the
2017.3 version. In order to maintain ABI, that's where they have to
stay.

https://phabricator.endlessm.com/T17204


pwithnall:
Is this still needed? I thought we’d rebuilt every package which used the new ABI so that it was built against OSTree 2017.6.

pwithnall:
I don’t think it is still needed. Please reopen the PR if I’m wrong.

PR #74, opened 2017-07-11
Update to OSTree 2017.7

pwithnall:
Rebase our patchset on the upstream 2017.7 release. The `get_commit_sizes()` patches had to be reworked a little due to changes in how libostree writes objects (upstream has split the code paths for writing content and metadata). A few test updates and some minor fixes too. All noted in the commit messages as ‘2017.7’.

This should eventually be pushed with the `Version_2017.7` tag (updated).

The packaging side of this is PR #75.


src/libostree/ostree-repo-commit.c
```diff
@@ -1867,7 +2033,8 @@ ostree_repo_write_content (OstreeRepo       *self,
                                           OSTREE_OBJECT_TYPE_FILE, &have_obj,
                                           cancellable, error))
         return FALSE;
-      if (have_obj)
+      if (have_obj &&
+          !(ostree_repo_get_mode (self) == OSTREE_REPO_MODE_ARCHIVE_Z2 && self->generate_sizes))
```
dbnicholson: This seems like the right thing to do, but actually a change from the old behavior. In the current single `write_object()` case, it seems that [having the correct object](https://github.com/endlessm/ostree/blob/da533f23c58fa8ed694456952ecef5d09bf95d37/src/libostree/ostree-repo-commit.c#L789) always returns early before `repo_store_size_entry`.

Actually, now that I look at it more, it seems like `ostree_repo_write_content` and `ostree_repo_write_metadata` are always called with a `NULL` expected checksum when creating commits, so you'd never reach here. On the other hand, checking `self->generate_sizes` seems like a better safeguard. In which case, I think this same condition should be added to `ostree_repo_write_metadata`.

dbnicholson:
> This should eventually be pushed with the Version_2017.7-1-endless1 tag.

I don't think that's right. We normally tag the upstream version so that the calculated version matches the number of commits since the upstream release. I.e., you should apply tag `Version_2017.7` to upstream's `v2017.7` tag.

src/libostree/ostree-repo-commit.c
```diff
@@ -1859,6 +2022,9 @@ ostree_repo_write_content (OstreeRepo       *self,
 {
   /* First, if we have an expected checksum, see if we already have this
    * object.  This mirrors the same logic in ostree_repo_write_metadata().
+   *
+   * If we’re writing out size metadata, fall through to the write_content_object()
+   * call anyway, since we need all the objects to end up in self->object_sizes.
```
pwithnall: Ah yes, it does. I’ll fix that.

pwithnall:
Thanks for the review. I’ve updated the branch to address the review issues and [kicked off another OBS build](https://obs-master.endlessm-sf.com/package/show?package=ostree&project=home%3Awithnall%3Abranches%3Aeos%3Amaster%3Aendless). I’ll check the status of that in the morning.

pwithnall:
The build succeeded and has passed some basic smoketesting with `ostree` and gnome-software. I’ve [opened an OBS request](https://obs-master.endlessm-sf.com/request/show/13757).

src/libostree/ostree-repo-commit.c
dbnicholson: Yeah, looks good. BTW, if you do upstream this, it would be nice to try to make the condition simpler. Like, only set `generate_sizes` if the mode is archive-z2. Then you can just check that and don't also need the separate `indexable` variable.

Do you know why the sizes are only generated on archive-z2? I didn't work on this feature and never noticed it until this review.

dbnicholson:
I declined the OBS request. We just need to get the branches and tags pushed and then jenkins will pick it up and make the packages for OBS. (That's the whole point of testing the build with jenkins-obs)

PR #75, opened 2017-07-11
debian: Update packaging to Debian 2017.7-1 release

pwithnall:
- Keep Endless specific ostree_repo_delete_compat_signature() and
   ostree_repo_get_commit_sizes() ABIs.
 - Keep the parts needed for boot in the main ostree package, as
   ostree-boot does not yet exist in the upstream Debian packaging.
 - Keep dependency on libgpgme11-dev rather than libgpgme-dev as in
   upstream packaging, as EOS still has older gpgme packaging.

---

Rebase our patchset on the upstream 2017.7 release. No major changes to the packaging apart from the `libgpgme11-dev` dependency change noted above.

This should eventually be pushed with the `Version_2017.7_debian` tag (updated).

The code side of this is PR #74.


debian/changelog
```diff
@@ -1,3 +1,27 @@
+ostree (2017.7-1-endless1) unstable; urgency=medium
```
dbnicholson: I don't think this version is what you want. Having multiple `-` will make the upstream version `2017.7-1`.
```
$ python3
>>> from debian.debian_support import Version
>>> v = Version('2017.7-1-endless1')
>>> v.upstream_version
'2017.7-1'
>>> v.debian_revision
'endless1'
```
Normally there's no hyphen between the debian revision and the downstream version. E.g., `1endless1`.

This also could cause issues with the version calculation in `jenkins-obs-build`. It expects (incorrectly) that you'll start the debian revision at 0 and any subsequent commits on the debian branch will increase that. However, since the upstream branch has many commits after the upstream tag, the upstream version will be newer than what you have here and it won't matter.

So, I think you can leave this as `2017.7-1endless1` since we know it won't be an issue here. Eventually I will make it so that the versions can be more flexible...

debian/control
```diff
@@ -14,23 +14,23 @@ Build-Depends:
...
- libc6-dev (>= 2.19-18+deb8u4endless4) [!amd64 !i386],
  libcap-dev,
  libfuse-dev,
- libgirepository1.0-dev (>= 1.34.0),
+ libgirepository1.0-dev,
  libglib2.0-dev (>= 2.40.0),
- libgpgme11-dev (>= 1.1.8),
- liblzma-dev (>= 5.0.5),
+ libgpgme11-dev,
+ liblzma-dev,
```
dbnicholson: I guess it's fine to drop all the version requirements since we likely have them on master and eos3.2.

debian/control
```diff
@@ -53,9 +53,6 @@ Section: introspection
 Depends:
  ${gir:Depends},
  ${misc:Depends},
-Provides: gir1.2-ostree
-Conflicts: gir1.2-ostree
-Breaks: gir1.2-ostree
```
dbnicholson: I suppose there's no harm in removing this and the libostree-1 back compat names. It's been long enough.

debian/changelog
pwithnall: OK, I’ll try `2017.7-1endless1`. One of these days I will remember the syntax for Debian package versioning.

pwithnall: ⇒ This will become the `Version_2017.7_debian` tag, right? (Not `Version_2017.7-1endless1_debian` — `jenkins-obs` complains about that. Specifically, it seems to unconditionally look for the `Version_2017.7_debian` tag.)

pwithnall:
Thanks for the review. I’ve updated the branch to address the review issues and [kicked off another OBS build](https://obs-master.endlessm-sf.com/package/show?package=ostree&project=home%3Awithnall%3Abranches%3Aeos%3Amaster%3Aendless). I’ll check the status of that in the morning.

pwithnall:
The build succeeded and has passed some basic smoketesting with `ostree` and gnome-software. I’ve [opened an OBS request](https://obs-master.endlessm-sf.com/request/show/13757).

dbnicholson:
Yes, it should be `Version_2017.7_debian`. Once you make the tag on the main branch `Version_something`, then `jenkins-obs-build` will look for `Version_something_debian` on the debian branch.

You _should_ be able to tag the debian branch in a more useful way to indicate (for instance) that you're at debian revision 2 or whatever. In my local branch, it allows a bunch of that and consults debian/changelog to make sure they're in agreement. But it's not quite polished off.

BTW, I cancelled the OBS request. We just need to get the branches and tags pushed and then jenkins will pick it up and make the packages for OBS. (That's the whole point of testing the build with `jenkins-obs`)

PR #78, opened 2017-08-16
debian: Drop <!nodoc> profile from ostree.install due to lack of tooling

pwithnall:
Our toolchain doesn’t currently support this profile, which results in
the man pages getting installed in `/<!nodoc>`, which is suboptimal.

https://phabricator.endlessm.com/T18832


cosimoc:
LGTM (we may want to open another ticket to add the tooling)

pwithnall:
Filed https://phabricator.endlessm.com/T18835.

PR #79, opened 2017-08-16
debian: Drop redundant ostree-boot.* files

pwithnall:
These hold redundant lists of files which are already listed in
ostree.install and friends, and are from an abortive split of the ostree
package into ostree and ostree-boot. We may still do this at some point
in the future, but not right now. Keeping the ostree-boot files around
for the moment is needlessly confusing for packagers.

https://phabricator.endlessm.com/T188836


pwithnall:
Done. I also filed https://phabricator.endlessm.com/T18848 as a follow-up task for splitting `ostree-boot` out again (properly).

PR #80, opened 2017-08-24
Update to OSTree 2017.10 and enable --enable-experimental-API (T16919)

pwithnall:
https://phabricator.endlessm.com/T16919

PR #81, opened 2017-08-24
debian: Update to 2017.10 and enable --enable-experimental-api flag

pwithnall:
* New upstream release
   - Install bash-completion script
 * Enable --enable-experimental-api configure flag — this will enable P2P
   functionality in libostree

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,13 @@
+ostree (2017.10-1endless1) unstable; urgency=medium
+
+  * New upstream release
+    - Update symbols file
+    - Add Build-Depends-Indep for docs as per upstream packaging
+  * Enable --enable-experimental-api configure flag — this will enable P2P
+    functionality in libostree
+
+ -- Philip Withnall <withnall@endlessm.com>  Wed, 22 Aug 2017 12:08:00 +0100
+
 ostree (2017.7-1endless1) unstable; urgency=medium
 
   * Resync Debian packaging changes
--- a/debian/control
+++ b/debian/control
@@ -41,6 +41,8 @@ Build-Depends:
  python-yaml <!nocheck>,
  xsltproc <!nodoc>,
  zlib1g-dev,
+Build-Depends-Indep:
+ libglib2.0-doc,
 Standards-Version: 4.0.0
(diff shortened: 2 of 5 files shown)


pwithnall:
(Code changes in PR #80.)

debian/control
```diff
@@ -41,6 +41,8 @@ Build-Depends:
  python-yaml <!nocheck>,
  xsltproc <!nodoc>,
  zlib1g-dev,
+Build-Depends-Indep:
+ libglib2.0-doc,
```
dbnicholson: I believe you here, but how is this enforced upstream?

debian/rules
```diff
@@ -25,6 +26,7 @@ configure_options = \
 	--with-grub2-mkconfig-path=/usr/sbin/grub-mkconfig \
 	--with-systemdsystemgeneratordir=/lib/systemd/system-generators \
 	--with-systemdsystemunitdir=/lib/systemd/system \
+	BASH_COMPLETIONSDIR=/usr/share/bash-completion/completions \
```
dbnicholson: Any reason not to rely on pkg-config here?

debian/changelog
```diff
@@ -1,3 +1,13 @@
+ostree (2017.10-1endless1) unstable; urgency=medium
```
dbnicholson: You may want to use `-0endless1` here in case debian catches up and we want to incorporate their `-1` version. Kinda theoretical since jenkins is going to make up a much newer upstream version, though...

cosimoc:
We just pushed this with the change Dan suggested to use `-0endless1` instead. Closing this.

debian/rules
pwithnall: IIRC, for the same reason we don’t rely on it for the `--with-systemd*` arguments — it can end up retrieving a path in the wrong prefix.

debian/control
pwithnall: I’m not quite sure what you mean. How is *what* enforced upstream?

dbnicholson: I figured this out later. I had assumed this was added to satisfy some build check. But actually the reason to have it is so that the generated docs get cross references to glib's docs.

PR #82, opened 2017-09-11
Fix handling of /etc/ostree/remotes.d

dbnicholson:
This is a backport of https://github.com/ostreedev/ostree/pull/1151 and https://github.com/ostreedev/ostree/pull/1155 that allows fixing how /etc/ostree/remotes.d is used when remotes are added. This was breaking 2 cases we care about at Endless:

1. Flatpak using system ostree repo
2. ostree-receive service specifying a remotes config dir for a non-system repo

https://phabricator.endlessm.com/T19077

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -122,6 +122,27 @@ Boston, MA 02111-1307, USA.
         keep free. The default value is 3.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>add-remotes-config-dir</varname></term>
+        <listitem>
+          <para>
+            Boolean value controlling whether new remotes will be added
+            in the remotes configuration directory. Defaults to
+            <literal>true</literal> for system ostree repositories. When
+            this is <literal>false</literal>, remotes will be added in
+            the repository's <filename>config</filename> file.
+          </para>
+          <para>
+            This only applies to repositories that use a remotes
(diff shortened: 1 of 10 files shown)


man/ostree.repo-config.xml
```diff
@@ -122,6 +122,27 @@ Boston, MA 02111-1307, USA.
...
+            this is <literal>false</literal>, remotes will be added in
+            the repository's <filename>config</filename> file.
+          </para>
+          <para>
+            This only applies to repositories that use a remotes
+            configuration directory such as system ostree repositories,
+            which use <filename>/etc/ostree/remotes.d</filename>.
+            Non-system repositories do not use a remotes configuration
+            directory unless one is specified when the repository is
+            opened.
```
mwleeds: should say "created" I think

dbnicholson: Unfortunately, not true. You specify the remotes config dir as a GObject property when you open the repo. Or, for a system repo, it's magically set to /etc/ostree/remotes.d.

It might be good to be able to specify the remotes config dir with a repo config option, but that doesn't exist today.

dbnicholson: Maybe "initialized" since it's at object initialization time (whether the repo exists yet or not) that you specify the property. But I don't think initialized would mean a lot to the reader of this documentation.

mwleeds: ah, ok

PR #84, opened 2017-10-03
debian: Update to 2017.12 and re-sync Debian packaging

pwithnall:
* Resync Debian packaging changes
   - Resync to packaging from 2017.11-2
   - Keep differences as per README.Endless
   - Drop all debian/patches/ as they’re all present in the 2017.12 release
   - Keep <!nodoc> profile usage from Debian packaging, as T18835 is now fixed
   - Keep [ -e Makefile ] guard in dh_auto_clean to avoid build failures
     from a clean source tree
 * New upstream release
   - Update symbols file
   - Rebase Endless patches

https://phabricator.endlessm.com/T18936

---

The corresponding code changes are in PR #85.


debian/rules
```diff
@@ -62,8 +63,12 @@ override_dh_install:
 	:
 	rm -f debian/tmp/etc/grub.d/15_ostree
 	rm -f debian/tmp/usr/lib/libostree/grub2-15_ostree
+	rm -f debian/tmp/usr/lib/tmpfiles.d/ostree-tmpfiles.conf
```
dbnicholson: I think we may actually want to ship this. All it does is cleanup the left behind overlays at boot from `ostree admin unlock` (non-hotfix mode).

pwithnall: Fix pushed.

debian/rules
```diff
@@ -40,11 +38,14 @@ endif
...
 
-override_dh_auto_test:
+override_dh_auto_test-arch:
 ifeq (,$(filter nocheck,$(DEB_BUILD_OPTIONS)))
 	debian/test.sh
 endif
 
+override_dh_auto_test-indep:
+	:
+
```
smspillaz: Hmm, what does this do?

PR #87, opened 2017-10-24
Various OSTree bug fixes

pwithnall:
Cherry-picked from upstream (they will all be in the upcoming 2017.13 release):
 * https://github.com/ostreedev/ostree/pull/1281
 * https://github.com/ostreedev/ostree/pull/1293
 * https://github.com/ostreedev/ostree/pull/1301
 * https://github.com/ostreedev/ostree/pull/1303

---

https://phabricator.endlessm.com/T19293

PR #89, opened 2017-10-24
Update server ostree to 2017.12

dbnicholson:
This syncs the xenial branch to master for putting 2017.12 on the server. Please review but don't merge (not that you can merge this, anyways).

Debian packaging in #90.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -4132,7 +4132,9 @@ ostree_repo_sign_commit (OstreeRepo     *self,
    * pass the homedir so that the signing key can be imported, allowing
    * subkey signatures to be recognised. */
   g_autoptr(GError) local_error = NULL;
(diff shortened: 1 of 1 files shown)


pwithnall:
I’m going to totally ignore what the github interface splurges everywhere, and review this one via [`git lola`](http://blog.kfish.org/2010/04/git-lola.html), which shows me that this brings `T18830-infra-update` to the same point in the commit graph as `master`.

That means it’s missing the commits from #87 which were just pulled, but it should be good otherwise. Looks good to push if you additionally pull and push those commits.

dbnicholson:
> That means it’s missing the commits from #87 which were just pulled, but it should be good otherwise. Looks good to push if you additionally pull and push those commits.

I'll update with your latest changes now.

dbnicholson:
I pulled in those 2 PRs from master and am going to go ahead and push now.

PR #88, opened 2017-10-24
debian: Add new symbols for OstreeRepoFinderOverride

pwithnall:
Signed-off-by: Philip Withnall <withnall@endlessm.com>

https://phabricator.endlessm.com/T19293

See https://github.com/endlessm/ostree/pull/87 for the non-Debian part of this.

PR #90, opened 2017-10-24
Update server ostree to 2017.12 - debian packaging

dbnicholson:
Update the xenial packaging to 2017.12. Note that this is based on debian's master branch in https://anonscm.debian.org/cgit/collab-maint/ostree.git. They have the full history from upstream, so the relevant commits are just the last few. I'd like to move to this model when debian has git history as it makes the rebase/merge much easier.

Please review but don't merge as I'll push along with the tags. Main branch review is in #89.


pwithnall:
> I'd like to move to this model when debian has git history as it makes the rebase/merge much easier.

Yes please!

> so the relevant commits are just the last few

i.e. The topmost 3, plus the merge commit (number 4)?

Everything looks OK to me apart from the fact that the new `OstreeRepoFinderOverride` symbols from #88 need to be added.

dbnicholson:
>> so the relevant commits are just the last few

> i.e. The topmost 3, plus the merge commit (number 4)?

Really everything past the upstream "Release 2017.12" commit, but yeah, my commits are just the merge (which I did to keep my previous xenial backport fixes) + 3 on top.

> Everything looks OK to me apart from the fact that the new OstreeRepoFinderOverride symbols from #88 need to be added.

I'll pull that in now.

dbnicholson:
I pushed this on debian-xenial.

PR #93, opened 2017-11-08
tests: Don't symlink rofiles-fuse if it's disabled

dbnicholson:
Creating the symlink will cause make to try to build rofiles-fuse, which
will fail if it's disabled. Normally I wouldn't disable rofiles-fuse,
but it's triggering a hang in our ARM Xenial builder's kernel in splice.
I'm sure that's fuse's fault, but for now I just need to disable
rofiles-fuse there and found --disable-rofiles-fuse didn't actually
work.

Closes: #1325
Approved by: cgwalters

This is a backport of https://github.com/ostreedev/ostree/commit/2e9fea8fa54f431a6e4cc2993d0b4b2ca18a6070.

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -41,7 +41,7 @@ AM_TESTS_ENVIRONMENT += OT_SKIP_READDIR_RAND=1 G_SLICE=always-malloc
 endif
 
 uninstalled_test_data = tests/ostree-symlink-stamp tests/ostree-prepare-root-symlink-stamp \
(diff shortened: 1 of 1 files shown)


dbnicholson:
You got it. The tests need the program to be in the tests directory, so there's a pattern rule that makes a symlink from the program in the top directory into the tests directory. However, the program itself is a prerequisite of the symlink, so it tries to build rofiles-fuse. That fails because the sources are under the `BUILDOPT_FUSE` conditional.

PR #95, opened 2017-11-14
T19126 GIR fixes

pwithnall:
Backporting a few fixes which have already been accepted upstream, to improve the GIR file for OSTree with experimental API enabled.

### DIFF
--- a/src/libostree/libostree-experimental.sym
+++ b/src/libostree/libostree-experimental.sym
@@ -89,3 +89,8 @@ global:
   ostree_repo_finder_override_get_type;
   ostree_repo_finder_override_new;
 } LIBOSTREE_2017.12_EXPERIMENTAL;
+
+LIBOSTREE_2017.14_EXPERIMENTAL {
+global:
+  ostree_remote_get_type;
+} LIBOSTREE_2017.13_EXPERIMENTAL;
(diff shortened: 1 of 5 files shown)


src/libostree/libostree-experimental.sym
```diff
@@ -89,3 +89,8 @@ global:
   ostree_repo_finder_override_get_type;
   ostree_repo_finder_override_new;
 } LIBOSTREE_2017.12_EXPERIMENTAL;
+
+LIBOSTREE_2017.14_EXPERIMENTAL {
+global:
+  ostree_remote_get_type;
+} LIBOSTREE_2017.13_EXPERIMENTAL;
```
dbnicholson: I hope it's not problem that we start defining this symbol version in our 2017.13 build.

dbnicholson:
By the way, if you're planning on using these in eos-build, you'll need them on the xenial branch, too.

src/libostree/libostree-experimental.sym
pwithnall: I don’t think it will be, since the identifiers in the `.sym` file don’t relate to the autoconf package version at all.

pwithnall: It does have to be listed in the Debian `.symbols` file though.

pwithnall:
> By the way, if you're planning on using these in eos-build, you'll need them on the xenial branch, too.

⇒ https://github.com/endlessm/ostree/pull/96

src/libostree/libostree-experimental.sym
pwithnall: ⇒ https://github.com/endlessm/ostree/pull/97

PR #103, opened 2017-12-19
Simplify compat object file handling

dbnicholson:
As described in the commit messages we no longer have any need for our legacy custom sig and sizes2 objects. Drop all handling of these files except the ability to prune them so that we don't leave dangling objects around.

https://phabricator.endlessm.com/T17367


dbnicholson:
See #104 for debian branch PR.

tests/test-compat-files.sh
```diff
@@ -71,9 +71,9 @@ mkdir local
 ${CMD_PREFIX} ostree --repo=local init --mode=archive-z2
 ${CMD_PREFIX} ostree --repo=local pull-local compat-repo main
 find local/objects -name '*.sig' | wc -l > sigcount
-assert_file_has_content sigcount "^1$"
+assert_file_has_content sigcount "^0$"
```
pwithnall: These two hunks aren’t in the original ‘repo: Copy compat metadata for pull-local’ commit, so we need to be careful to not drop them when we next rebase (I would have been tempted to just drop the original and ‘revert’ commits when rebasing, rather than squashing them and checking the result).

dbnicholson: I realized that after I sent the PR and should have said something. All of the changes to `tests/test-compat-files.sh` will need to be folded in on rebase (the result would definitely fail if not). It should be as simple as cherry picking `tests: Add tests for Endless compat files`, checking out the previous master's `tests/test-compat-files.sh` and amending the commit.

PR #106, opened 2018-01-09
debian: Compile with avahi support

mwleeds:
Add avahi-glib as a build dependency and use the configure option
--with-avahi to try to ensure that ostree compiles with avahi support,
which is necessary for P2P updates to work.

https://phabricator.endlessm.com/T20662

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -25,6 +25,8 @@ Build-Depends:
  gtk-doc-tools <!nodoc>,
  libarchive-dev,
  libattr1-dev,
+ libavahi-client-dev,
+ libavahi-glib-dev,
  libcap-dev,
  libfuse-dev,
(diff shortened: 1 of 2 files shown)


debian/control
```diff
@@ -9,6 +9,7 @@ Uploaders:
 Build-Depends:
  attr,
  autoconf,
+ avahi-glib,
```
dbnicholson: Is this right? I'd expect it to be libavahi-glib-dev or something like that. It has to be the package that contains the avahi-glib.pc file, I believe.

mwleeds: I just checked the contents of that package and you're right, it has `avahi-glib.pc`

mwleeds:
@ptomato thanks. Not sure what made me think it already had that dependency

PR #108, opened 2018-01-15
avahi: Don't complain with g_warning if the daemon wasn't running

smspillaz:
This is a normal case when running unit tests in client code
on continuous integration infrastructure. When those tests are
running they will set G_DEBUG=fatal-warnings which will cause
the program to abort if a warning is emitted. Instead, emit
a debug message if the problem was that we couldn't connect to
the daemon.

https://phabricator.endlessm.com/T16682

### DIFF
--- a/src/libostree/ostree-repo-finder-avahi.c
+++ b/src/libostree/ostree-repo-finder-avahi.c
@@ -1435,9 +1435,15 @@ ostree_repo_finder_avahi_start (OstreeRepoFinderAvahi  *self,
 
   if (client == NULL)
     {
-      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
-                   "Failed to create finder client: %s",
-                   avahi_strerror (failure));
+      if (failure == AVAHI_ERR_NO_DAEMON)
+        g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
+                     "Avahi daemon is not running: %s",
+                     avahi_strerror (failure));
+      else
+        g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                     "Failed to create finder client: %s",
+                     avahi_strerror (failure));
+
       return;
     }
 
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-repo-finder-avahi.c
```diff
@@ -1435,6 +1435,14 @@ ostree_repo_finder_avahi_start (OstreeRepoFinderAvahi  *self,
 
   if (client == NULL)
     {
+      if (failure == AVAHI_ERR_NO_DAEMON)
+        {
+          g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
+                       "Avahi daemon is not running: %s",
+                       avahi_strerror (failure));
+          return;
```
pwithnall: Nitpick: These two error paths could be simplified as:
```
if (failure == AVAHI_ERR_NO_DAEMON)
  g_set_error (…, G_IO_ERROR_NOT_FOUND, …);
else
  g_set_error (…, G_IO_ERROR_FAILED, …);
return;
```

src/libostree/ostree-repo-pull.c
```diff
@@ -4592,7 +4592,21 @@ ostree_repo_find_remotes_async (OstreeRepo                     *self,
...
-          g_warning ("Avahi finder failed; removing it: %s", local_error->message);
+          /* See ostree-repo-finder-avahi.c:ostree_repo_finder_avahi_start, we
+           * intentionally throw this so as to distinguish between the Avahi
+           * finder failing because the Avahi daemon wasn't running and
+           * the Avahi finder failing because of some actual error.
+           *
+           * We need to distinguish between g_debug and g_warning here because
+           * unit tests that use this code may set G_DEBUG=fatal-warnings which
+           * would cause client code to abort if a warning were emitted. See
+           * https://phabricator.endlessm.com/T16682#530304
```
pwithnall: (I realise this is not what I normally say, but) You can omit the Endless Phabricator link from this comment, since it won’t be useful to anyone upstream, and the comment explains things fairly well by itself.

pwithnall:
This still needs to be pushed upstream. @smspillaz, do you have time to do that, or shall I handle it?

smspillaz:
@pwithnall You are more than welcome to do so yourself if time permits - I am AFK at the moment

pwithnall:
Upstreamed as https://github.com/ostreedev/ostree/pull/1542, thanks.

PR #112, opened 2018-02-20
lib/pull: Allow specific commits in P2P updates

mwleeds:
Currently users of the find_remotes_async()/pull_from_remotes_async()
functions have no way to specify a commit hash to use instead of the
latest one available. This commit implements an "override-commit-ids"
option analogous to the one used by ostree_repo_pull_with_options().
It's accomplished by returning OstreeRepoFinderResult objects pointing
to the given commit checksum(s) regardless of which ones were available
from the remotes, but in the future this implementation could be
improved to take into account the commits advertised by the remotes.

One effect of this is that flatpak will have the ability to downgrade
apps that use collection IDs
(https://github.com/flatpak/flatpak/issues/1309).

Closes: #1425
Approved by: pwithnall

(cherry picked from commit 660bc588ee62717dbccaca18b6d2c5e4120c0901)

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -4524,7 +4524,11 @@ static void find_remotes_cb (GObject      *obj,
  * Pass the results to ostree_repo_pull_from_remotes_async() to pull the given @refs
  * from those remotes.
  *
- * No @options are currently supported.
+ * The following @options are currently defined:
+ *
+ *   * `override-commit-ids` (`as`): Array of specific commit IDs to fetch. The nth
+ *   commit ID applies to the nth ref, so this must be the same length as @refs, if
+ *   provided.
  *
  * @finders must be a non-empty %NULL-terminated array of the #OstreeRepoFinder
  * instances to use, or %NULL to use the system default set of finders, which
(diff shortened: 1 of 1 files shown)


pwithnall:
You might want to make a Debian package release for endlessm/ostree now, so that you can make other packages depend on it, so they are guaranteed to get the new `override-commit-ids` API. Failure to use an appropriate OSTree version would be a bit insidious, since it’s a soft API and its absence would result in a runtime failure rather than a linking failure.

mwleeds:
I'm not sure how to do that, but if we cherry-pick this and the flatpak PR to eos3.3 it would be difficult for someone to use the wrong version. And I think the behavior if you use the updated flatpak without the updated ostree is the same as if you update neither.

mwleeds:
@rshuler could you or I cherry-pick this and the flatpak PR? It would make testing LAN/USB stuff easier.

rshuler:
@mwleeds Cherry-picked this and the flatpak PR to eos3.3, and delayed tonight's builds by an hour to give them time to build so they will be QA ready 2/22.  Normally, I prefer to have a specific ticket in phabricator so we can clearly track what is changing on a release, and then if I see the ticket in QA blocked I know that it needs to be cherry-picked.

PR #114, opened 2018-04-16
Rebase on 2018.4

dbnicholson:
Comparison branch for rebasing on 2018.4. Notable changes here:

* All the compat object stuff done in https://phabricator.endlessm.com/T17367 has been squashed down to 2 commits
* The `size-summary` builtin command needed a little adjusting for how ostree handles the command line args now
* We still have an upstreamable (I believe) change about avahi errors vs warnings
* Changed the repo locking to be enabled by default since upstream disabled it by default

This branch is for review only. After comments have been addressed, I'll push the branches and tags manually.

https://phabricator.endlessm.com/T22022

PR #115, opened 2018-04-16
Sync 2018.4-1 debian packaging

dbnicholson:
Comparison branch for pulling in 2018.4-1 debian packaging. Main branch changes are in #114. From the commit message:

- Drop package version back to 0 to appease jenkins
- Resync to packaging from 2018.4-1
- Keep differences as per README.Endless
- Keep experimental API options --with-experimental-api and --with-avahi
- Keep --enable-trivial-httpd-cmdline option as I'm not sure if we have
  code that still expects this
- Drop BASH_COMPLETIONSDIR from configure options as the build seems to
  pick that up appropriately
- Drop GJS override to disable gjs tests since we don't have issues
  installing gjs like debian does
- Drop dh_auto_clean override for Makefile existence since that doesn't
  seem to be a problem anymore

This PR is for review only. After the comments have been addressed, I'll push the branches and tags manually.

https://phabricator.endlessm.com/T22022

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,20 @@
+ostree (2018.4-2endless1) master; urgency=medium
+
+  * Resync Debian packaging changes (T22022):
+    - Rewrite history on debian/2018.4-2 tag
+    - Reapply changes from README.Endless
+    - Add --enable-experimental-api and all symbols
+    - Update versions in symbols files to use upstream versions instead of
+      our downstream package versions
+    - Drop Don-t-write-to-parent-repo.patch as it's on our main branch
+    - Revert change to disable gjs tests as we don't have the install
+      issue on Endless
+    - Add back --enable-trivial-httpd-cmdline as eos-updater uses that in
+      its tests
+    - Add --with-avahi and associated dependencies for P2P support
+
+ -- Dan Nicholson <nicholson@endlessm.com>  Wed, 18 Apr 2018 09:40:33 -0500
+
 ostree (2018.4-2) unstable; urgency=medium
 
   * d/p/Don-t-write-to-parent-repo.patch:


dbnicholson:
retest this please

debian/changelog
```diff
@@ -1,3 +1,155 @@
+ostree (2018.4-0) master; urgency=medium
+
+  * Resync Debian packaging changes (T22022):
+    - Drop package version back to 0 to appease jenkins
```
pwithnall: Couldn’t we version it `2018.4-1endless0`? Having decrementing version numbers in the `debian/changelog` seems a bit wrong.

debian/rules
```diff
@@ -41,6 +42,7 @@ override_dh_auto_configure:
 
 override_dh_auto_test-arch:
 ifeq (,$(filter nocheck,$(DEB_BUILD_OPTIONS)))
+	chmod +x tests/test-pull-bareuseronly.sh
```
pwithnall: Why is this needed? `test-pull-bareuseronly.sh` is mode 0755 in git, and this apparently isn’t needed for any of the other tests.

debian/changelog
dbnicholson: Sure. `jenkins-obs-build` will start the packaging version at 0, which is why I have that here. However, since we have patches on the main branch, the upstream version will move past this, so it won't matter. See https://phabricator.endlessm.com/w/software/build/ci-integration/#version-numbering.

debian/rules
dbnicholson: This came from debian. Possibly it's not 0755 in the `make dist` tarball. Let me check.

dbnicholson: That looks like it can be dropped now - https://salsa.debian.org/debian/ostree/commit/022fab56dc536c07667e89d63f836ad05c013af7. He'd added it through a patch.

dbnicholson:
Simon just pushed out 2018.4-2 with this https://salsa.debian.org/debian/ostree/commit/76cbbc1a69d40f7664b1f77638ef4f17e878d9f6, which seems kinda important...

dbnicholson:
@pwithnall Since I'm likely going to change the packaging here so it's based on 2018.4-2, do you mind if I change our debian branch into a rebase on https://salsa.debian.org/debian/ostree? I've been doing that on the debian-xenial branch and I find it so much more usable than just blindly copying in files to the debian directory and trying to figure out what changed.

pwithnall:
> @pwithnall Since I'm likely going to change the packaging here so it's based on 2018.4-2, do you mind if I change our debian branch into a rebase on https://salsa.debian.org/debian/ostree? I've been doing that on the debian-xenial branch and I find it so much more usable than just blindly copying in files to the debian directory and trying to figure out what changed.

**100% agree**, that would make me so happy :confetti_ball: :fireworks: :champagne:

dbnicholson:
PR rebuilt so that the debian-master branch is a descendent of https://salsa.debian.org/debian/ostree/tags/debian%2F2018.4-2. I then cherry-picked/squashed some of our own commits on top so that the end result was what we wanted without a ton of noise. Should be much easier to manage on the next rebase/merge. Of course, debian keeps the entire main branch in they're history, so this will need to be force pushed.

This built for me in https://obs-master.endlessm-sf.com/package/show?package=ostree&project=home%3Adbn%3Abranches%3Aeos%3Amaster%3Aendless. The build failure here is due to https://github.com/endlessm/eos-build/pull/773.

dbnicholson:
retest this please

pwithnall:
> That said, I don't feel too strongly about it if you want to drop that patch.

Up to you.

dbnicholson:
Updated. I added a pointer to debian's git in README.Endless and mentioned the symbols file version changes in the changelog. I decided to keep the old changelog entries since I think it's kinda nice to look back and see how we've handled things.

dbnicholson:
BTW, a more condensed view of our rebase can be seen at https://github.com/endlessm/ostree/compare/debian/2018.4-2...T22022-update-2018.4-debian.

dbnicholson:
Force pushed to debian-master now.

PR #122, opened 2018-06-06
lib/repo-pull: Retry pulls without static deltas if they fail

pwithnall:
If pulling a static delta fails (due to one part of it missing from a
mirror, for example), retry the entire pull operation with
disable-static-deltas=true.

This is a functional, but ugly approach to fixing the problem, and a
neater solution may be adopted upstream in future. If so, this commit
should be dropped and replaced with the upstream solution:

https://github.com/ostreedev/ostree/pull/1612

https://phabricator.endlessm.com/T22873

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -3417,13 +3417,13 @@ initiate_request (OtPullData                 *pull_data,
  *     a transient network error, such as a socket timeout; default is 5, 0
  *     means return errors without retrying
  */
-gboolean
-ostree_repo_pull_with_options (OstreeRepo             *self,
-                               const char             *remote_name_or_baseurl,
-                               GVariant               *options,
-                               OstreeAsyncProgress    *progress,
-                               GCancellable           *cancellable,
-                               GError                **error)
+static gboolean
+ostree_repo_pull_with_options_internal (OstreeRepo           *self,
+                                        const char           *remote_name_or_baseurl,
+                                        GVariant             *options,
+                                        OstreeAsyncProgress  *progress,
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -4541,6 +4541,43 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
...
+{
+  g_autoptr(GError) local_error = NULL;
+
+  if (!ostree_repo_pull_with_options_internal (self, remote_name_or_baseurl,
+                                               options, progress, cancellable,
+                                               &local_error))
+    {
+      if (g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND) &&
+          (options == NULL ||
+           g_variant_lookup_value (options, "disable-static-deltas", NULL) == NULL))
```
dbnicholson: What if you passed `disable-static-deltas` as `FALSE`? Seems like you should still get to use this fallback. Also, what about `require-static-deltas`? You shouldn't use the fallback if that's `TRUE`.

pwithnall: Good catches.

pwithnall: Fix force-pushed.

dbnicholson:
> FTR, I never said it was, and submitted it as a ‘first working attempt at a fix’ upstream to see if they were happy about the general idea of retrying failed deltas.

Yeah, sorry, I wrote that poorly. I saw that this was just a rough concept and was agreeing with the discussion upstream that there are better ways to accomplish this long term.

PR #124, opened 2018-06-14
lib/repo: Change default of core.add-remotes-config-dir to false

pwithnall:
In upstream OSTree, the default for core.add-remotes-config-dir is true
for system repositories and false for all other repositories. Since we
use the same repository for the OS and for flatpak, we need the
configuration value to be the same for both (otherwise there can be
conflicts between /etc/flatpak/remotes.d/*.conf and
/ostree/repo/config).

While we set core.add-remotes-config-dir=false in all our configurations
for OSTree, this patch exists for paranoia, as the consequences of using
the wrong default value for it are hard to recover from (essentially, a
non-updatable system).

See https://phabricator.endlessm.com/T22258.

This is a downstream patch which should not be upstreamed, and should
not be dropped from our fork of OSTree.

https://phabricator.endlessm.com/T22970

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -130,7 +130,7 @@ Boston, MA 02111-1307, USA.
           <para>
             Boolean value controlling whether new remotes will be added
             in the remotes configuration directory. Defaults to
-            <literal>true</literal> for system ostree repositories. When
+            <literal>false</literal> for all ostree repositories. When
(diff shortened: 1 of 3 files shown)


pwithnall:
Closed in favour of changing this in flatpak:
 * https://github.com/endlessm/flatpak/pull/99
 * https://phabricator.endlessm.com/T22970#604642

PR #126, opened 2018-06-28
Upstream cherry picks min-free-space-size

uajain:
https://phabricator.endlessm.com/T13698

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -668,6 +668,33 @@ _ostree_repo_bare_content_commit (OstreeRepo                 *self,
 {
   OstreeRealRepoBareContent *real = (OstreeRealRepoBareContent*) barewrite;
   g_assert (real->initialized);
+
+  if ((self->min_free_space_percent > 0 || self->min_free_space_mb > 0) && self->in_transaction)
+    {
+      struct stat st_buf;
+      if (!glnx_fstat (real->tmpf.fd, &st_buf, error))
+        return FALSE;
+
+      g_mutex_lock (&self->txn_lock);
+      g_assert_cmpint (self->txn.blocksize, >, 0);
+
+      const fsblkcnt_t object_blocks = (st_buf.st_size / self->txn.blocksize) + 1;
+      if (object_blocks > self->txn.max_blocks)
+        {
+          g_mutex_unlock (&self->txn_lock);
+          g_autofree char *formatted_required = g_format_size (st_buf.st_size);
+          if (self->min_free_space_percent > 0)
+           return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s more required",
+                              self->min_free_space_percent, formatted_required);
+          else
+            return glnx_throw (error, "min-free-space-size %luMB woulid be exceeded, %s more required",
+                               self->min_free_space_mb, formatted_required);
+        }
+      /* This is the main bit that needs mutex protection */
+      self->txn.max_blocks -= object_blocks;
+      g_mutex_unlock (&self->txn_lock);
+    }
+
   ot_checksum_get_hexdigest (&real->checksum, checksum_buf, buflen);
 
   if (real->expected_checksum &&
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1037,7 +1037,7 @@ write_content_object (OstreeRepo         *self,
     size = 0;
 
   /* Free space check; only applies during transactions */
-  if (self->min_free_space_percent > 0 && self->in_transaction)
+  if ((self->min_free_space_percent > 0 || self->min_free_space_size > 0) && self->in_transaction)
     {
       g_mutex_lock (&self->txn_lock);
       g_assert_cmpint (self->txn.blocksize, >, 0);
@@ -1046,8 +1046,12 @@ write_content_object (OstreeRepo         *self,
         {
           g_mutex_unlock (&self->txn_lock);
           g_autofree char *formatted_required = g_format_size ((guint64)object_blocks * self->txn.blocksize);
-          return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s more required",
-                             self->min_free_space_percent, formatted_required);
+          if (self->min_free_space_percent > 0)
+            return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s more required",
+                               self->min_free_space_percent, formatted_required);
+          else
+            return glnx_throw (error, "min-free-space-size %luMB would be exceeded, %s more required",
+                               self->min_free_space_size, formatted_required);
         }
       /* This is the main bit that needs mutex protection */
       self->txn.max_blocks -= object_blocks;
@@ -1656,6 +1660,25 @@ devino_cache_lookup (OstreeRepo           *self,
   return dev_ino_val->checksum;
 }
 
+static guint64
+min_free_space_calculate_reserved_blocks (OstreeRepo *self, struct statvfs *stvfsbuf)
+{
+  guint64 reserved_blocks = 0;
+
+  if (self->min_free_space_size > 0)
+    {
+      reserved_blocks = (self->min_free_space_size << 20) / stvfsbuf->f_bsize;
+    }
+  else if (self->min_free_space_percent > 0)
+    {
+      /* Convert fragment to blocks to compute the total */
+      guint64 total_blocks = (stvfsbuf->f_frsize * stvfsbuf->f_blocks) / stvfsbuf->f_bsize;
+      reserved_blocks = ((double)total_blocks) * (self->min_free_space_percent/100.0);
+    }
+
+  return reserved_blocks;
+}
+
 /**
  * ostree_repo_scan_hardlinks:
  * @self: An #OstreeRepo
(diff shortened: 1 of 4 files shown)


src/libostree/ostree-repo-commit.c
```diff
@@ -1738,26 +1791,31 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
     return FALSE;
 
   self->in_transaction = TRUE;
-  if (self->min_free_space_percent > 0)
+  self->cleanup_stagedir = FALSE;
+  if (self->min_free_space_percent >= 0 || self->min_free_space_mb >= 0)
```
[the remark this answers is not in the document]
uajain: `ostree_repo_prepare_transaction` is a special case: 
We need `>=` as that will lead to calculate the `reserved_blocks`. Therefore, if any one is mentioned to be `0` this will still be`TRUE`,  calculate `reserved_blocks`=0 and finally set `self->txn.max_block` (see below) accordingly.

And if you notice `self->txn.max_block` is how we decide we are over the  free-space quota or not. Once it is computed (in `prepare_transaction`) we can happily use `>` for `min-free-space-*` (which is what is happening). Does it make sense?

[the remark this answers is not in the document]
pwithnall: If you always need to take that branch, why have the `if` statement?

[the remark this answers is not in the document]
uajain: Point.

src/libostree/ostree-repo.c
```diff
@@ -2659,6 +2659,48 @@ get_remotes_d_dir (OstreeRepo          *self,
...
+        shifts = 10;
+        break;
+      case 'T':
+        shifts = 20;
+        break;
+      default:
+        g_assert_not_reached ();
+    }
+
+  self->min_free_space_mb = g_ascii_strtoull (size_str, NULL, 10) << shifts;
```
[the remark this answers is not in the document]
uajain: Right. I will fix the documentation.
OTOH, can you tell me what should be behavior in case there is an overflow or a `g_assert` will do the trick?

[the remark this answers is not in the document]
pwithnall: An overflow could be triggered by the user (because they write the config file), so has to be handled using a `GError`. `g_assert()` should only be used to check for programmer errors which can’t possibly be triggered by user input.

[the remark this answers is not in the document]
uajain: hmm, Do you mean whether to check for wrapping around? As `self->min_free_space_mb` is already `guint64` hence it won't "overflow" (rather wrap around). Not sure how to detect that? :thinking:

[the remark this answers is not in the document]
pwithnall: It will overflow, in the sense that if you want to fully represent a 64-bit integer shifted left by 20 bits, you need 84 bits of storage. You can only detect it before doing the operation:
```
guint64 min_free_space = g_ascii_strtoull (size_str, NULL, 10);
if (shifts > 0 && g_bit_nth_lsf (min_free_space, 64 - shifts) != -1)
  {
    g_set_error (…);
    return FALSE;
  }
self->min_free_space_mb = min_free_space << shifts;
```

That test goes by the principle that the shift will overflow if any of the `shifts` MSBs in `min_free_space` are set. I haven’t tested that, but it should be roughly correct.

uajain:
Whoops, conflicts! rebasing over master again.

src/libostree/ostree-repo-commit.c
```diff
@@ -668,6 +668,34 @@ _ostree_repo_bare_content_commit (OstreeRepo                 *self,
...
+      if (object_blocks > self->txn.max_blocks)
+        {
+          self->cleanup_stagedir = TRUE;
+          g_mutex_unlock (&self->txn_lock);
+          g_autofree char *formatted_required = g_format_size (st_buf.st_size);
+          if (self->min_free_space_percent > 0)
+           return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s more required",
+                              self->min_free_space_percent, formatted_required);
+          else
+            return glnx_throw (error, "min-free-space-size %luMB woulid be exceeded, %s more required",
```
pwithnall: Use `"%" G_GUINT64_FORMAT` rather than `"%lu"`, since the latter is not portable to other platforms (particularly 32-bit ones).

src/libostree/ostree-repo-commit.c
```diff
@@ -1037,17 +1065,23 @@ write_content_object (OstreeRepo         *self,
...
           g_mutex_unlock (&self->txn_lock);
-          g_autofree char *formatted_required = g_format_size ((guint64)object_blocks * self->txn.blocksize);
-          return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s more required",
-                             self->min_free_space_percent, formatted_required);
+          g_autofree char *formatted_required = g_format_size (bytes_required);
+          if (self->min_free_space_percent > 0)
+            return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s more required",
+                               self->min_free_space_percent, formatted_required);
+          else
+            return glnx_throw (error, "min-free-space-size %luMB would be exceeded, %s more required",
```
pwithnall: See above comment about `%lu`.

src/libostree/ostree-repo-commit.c
```diff
@@ -1744,30 +1797,32 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
...
+    {
+      guint64 bytes_required = bfree * self->txn.blocksize;
+      self->cleanup_stagedir = TRUE;
       g_mutex_unlock (&self->txn_lock);
+      g_autofree char *formatted_free = g_format_size (bytes_required);
+      if (self->min_free_space_percent > 0)
+        return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s available",
+                           self->min_free_space_percent, formatted_free);
+      else
+        return glnx_throw (error, "min-free-space-size %luMB would be exceeded, %s available",
```
pwithnall: See above comment about `%lu`.

src/libostree/ostree-repo-commit.c
```diff
@@ -668,6 +668,34 @@ _ostree_repo_bare_content_commit (OstreeRepo                 *self,
...
+
+      g_mutex_lock (&self->txn_lock);
+      g_assert_cmpint (self->txn.blocksize, >, 0);
+
+      const fsblkcnt_t object_blocks = (st_buf.st_size / self->txn.blocksize) + 1;
+      if (object_blocks > self->txn.max_blocks)
+        {
+          self->cleanup_stagedir = TRUE;
+          g_mutex_unlock (&self->txn_lock);
+          g_autofree char *formatted_required = g_format_size (st_buf.st_size);
```
pwithnall: This whole block which checks for whether there’s enough free space, and errors if there isn’t, could be factored out into a helper function to avoid it being repeated 3 times.

pwithnall: At the very least, the repetition of the same 2 error messages 3 times could be avoided.

src/libostree/ostree-repo.c
```diff
@@ -2659,6 +2659,52 @@ get_remotes_d_dir (OstreeRepo          *self,
...
+  if (g_once_init_enter (&regex_initialized))
+    {
+      regex = g_regex_new ("^([0-9]+)(G|M|T)B$", 0, 0, NULL);
+      g_assert (regex);
+      g_once_init_leave (&regex_initialized, 1);
+    }
+
+  g_autoptr(GMatchInfo) match = NULL;
+  if (!g_regex_match (regex, min_free_space_size_str, 0, &match))
+    return glnx_throw (error, "Failed to match '^[0-9]+[GMT]B$'");
```
pwithnall: Nitpick: That error message isn’t going to be very useful to the average user (most people don’t understand regexes), and it would end up bubbling up to gnome-software if someone edited their configuration. How about something like “min-free-space-size should be of the format ‘123TB’, ‘123GB’ or ‘123MB’”?

src/libostree/ostree-repo.c
```diff
@@ -2659,6 +2659,52 @@ get_remotes_d_dir (OstreeRepo          *self,
...
+      case 'T':
+        shifts = 20;
+        break;
+      default:
+        g_assert_not_reached ();
+    }
+
+  guint64 min_free_space = g_ascii_strtoull (size_str, NULL, 10);
+  if (shifts > 0 && g_bit_nth_lsf (min_free_space, 63 - shifts) != -1)
+    return glnx_throw (error, "Integer overflow detected");
```
pwithnall: Nitpick: Similarly, this error message could be more useful. How about “min-free-space-size value was too high”?

pwithnall: Though changing the error message here might mean the prefix message in `reload_core_config()` should be reworded.

src/libostree/ostree-repo-commit.c
pwithnall: Sure, something like the following would make me happy:
```
static gboolean
throw_min_free_space_error (OstreeRepo  *self,
                            guint64      bytes_required,
                            GError     **error)
{
  g_autofree char *formatted_free = g_format_size (bytes_required);
  if (self->min_free_space_percent > 0)
    return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s available",
                       self->min_free_space_percent, formatted_free);
  else
    return glnx_throw (error, "min-free-space-size %" G_GUINT64_FORMAT "MB would be exceeded, %s available",
                       self->min_free_space_mb, formatted_free);
}
```

src/libostree/ostree-repo.c
pwithnall: I don’t mind what the resulting error messages are, as long as they clearly point out to the user what the problem is with their configuration, and are reasonable English.

uajain:
@pwithnall Mind reviewing an upstream PR  too? 
* https://github.com/ostreedev/ostree/pull/1671

PR #127, opened 2018-07-02
lib/repo-commit: Abort a transaction if preparing it fails

pwithnall:
If ostree_repo_prepare_transaction() fails, we should reset the
repository’s state so that the failed call was essentially idempotent.
Do that by calling ostree_repo_abort_transaction() on the failure path.

Typically, the way for preparing a transaction to fail is for its
GCancellable to be triggered, rather than because any of the operations
involved in preparing a transaction are particularly failure prone.

Closes: #1647
Approved by: cgwalters

(Backported from upstream commit
abff8b8cfac00203062362ad506ffdb14f1da6ef.)

https://phabricator.endlessm.com/T23091


pwithnall:
autolockr

PR #129, opened 2018-07-19
Rebase to upstream 2018.6 (DON'T MERGE)

mwleeds:
This is the current state of upstream master plus 11 downstream commits we have to keep. None of the merge conflicts were very involved; see the commit messages for details. I still need to do the packaging branch.

https://phabricator.endlessm.com/T23138

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -4148,7 +4148,6 @@ ostree_repo_pull_with_options_internal (OstreeRepo           *self,
           {
             const char *delta;
             g_autoptr(GVariant) csum_v = NULL;
-            guchar *csum_data = g_malloc (OSTREE_SHA256_DIGEST_LEN);
             g_autoptr(GVariant) ref = g_variant_get_child_value (deltas, i);
 
             g_variant_get_child (ref, 0, "&s", &delta);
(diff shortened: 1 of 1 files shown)


mwleeds:
Here are my notes from the rebase in case they're helpful:
[ostree_2018.6_rebase.csv.txt](https://github.com/endlessm/ostree/files/2211721/ostree_2018.6_rebase.csv.txt)

src/libostree/ostree-repo-commit.c
```diff
@@ -1319,6 +1484,10 @@ write_metadata_object (OstreeRepo         *self,
        */
       if (have_obj)
         {
+          if (ostree_repo_get_mode (self) == OSTREE_REPO_MODE_ARCHIVE_Z2 && self->generate_sizes)
```
cosimoc: From https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-core.h#L189 it looks like the ARCHIVE_Z2 mode is just a legacy alias for the ARCHIVE mode. It may be worth changing it here to use the new value, and perhaps factor out this check into a small function, as it's repeated a few times in the patch.

mwleeds: Pushed a fixup to do this

mwleeds:
I rebased to get two more commits from upstream which meant dropping our downstream patch "Allow min-free-space-size and -percent to co-exist", and I edited the commit message and added a fixup for "Store sizes for all objects and add ostree_repo_get_commit_sizes" to address Cosimo's comment.

PR #130, opened 2018-07-20
Update packaging for 2018.6 rebase (DON'T MERGE)

mwleeds:


debian/ostree.install
```diff
@@ -1,8 +1,17 @@
 #!/usr/bin/dh-exec
 
+etc/dracut.conf.d/ostree.conf
 etc/ostree
+lib/systemd/system-generators/ostree-system-generator
+lib/systemd/system/ostree-finalize-staged.service
```
wjt: Will the presence of this unit cause us issues given we (presumably?) don't use staged deployment?

wjt:
Packaging looks fine, modulo the above query which isn't specifically about the packaging.

debian/ostree.install
wjt: I looked a bit closer. I *believe* what this new unit does is to defer the `/etc` merge to shutdown. (On upgrade, ostree compares `/etc` to `/usr/etc` in the booted deployment; any changed or created files are preserved in the new deployment's `/etc`, while any unmodified files are replaced with those from the new deployment's `/usr/etc`. I *think*.) So I guess it will be fine. I am more curious as to where the new deployment lives before this unit runs. Is it the same old place, and presumably after this service performs the merge it performs the `/boot/loader` swap to "enable" the new deployment? Or is it in some different staging location (in which case the eos-updater code that reads files from the new deployment might break)?

wjt: I don't fully understand https://github.com/ostreedev/ostree/blob/93da56842259f18bde23daf17dcefcf213b30b81/src/libostree/ostree-sysroot-deploy.c#L2808-L2885 but I think the behaviour is what I'd hope (the new, not-fully-prepared deployment is in the normal place). We'll really need to test eos-updater – specifically the post-upgrade flatpak installer – when upgrading from an eos version with this new libostree to an even newer eos version.

dbnicholson: I believe this is something we want and fixes https://phabricator.endlessm.com/T5658. As I understand it (haven't looked that closely at the implementation), what this does is delays the /etc merge until shutdown. Previously any changes made to /etc after the deployment was created would be lost. Upstream is https://github.com/ostreedev/ostree/issues/545 (and our old work in https://bugzilla.gnome.org/show_bug.cgi?id=758088).

I think we will want a little QA around this (attached to T5658) to make sure there aren't regressions introduced with the new units and how they might interact with our boot process.

mwleeds: I'll look into it a bit more. FWIW a relevant Endless bug is https://phabricator.endlessm.com/T5658

mwleeds: Since we're not using `ostree_sysroot_stage_tree` yet, this new service is a no-op. You can see in `_ostree_sysroot_finalize_staged` that it just returns TRUE if there aren't any staged deployments.

PR #131, opened 2018-07-23
config: Add a core/change-update-summary option

mwleeds:
This commits adds and implements a boolean repo config option called
"change-update-summary" which updates the summary file every time a ref
changes (additions, updates, and deletions).

The main impetus for this feature is that the `ostree create-usb` and
`flatpak create-usb` commands depend on the repo summary being up to
date. On the command line you can work around this by asking the user to
run `ostree summary --update` but in the case of GNOME Software calling
out to `flatpak create-usb` this wouldn't work because it's running as a
user and the repo is owned by root. That strategy also means flatpak
can't update the repo metadata refs for fear of invalidating the
summary.

Another use case for this relates to LAN updates. Specifically, the
component of eos-updater that generates DNS-SD records advertising ostree
refs depends on the repo summary being up to date.

Since ostree_repo_regenerate_summary() now takes an exclusive lock, this
should be safe to enable. However it's not enabled by default because of
the performance cost, and because it's more useful on clients than
servers (which likely have another mechanism for updating the summary).

Fixes https://github.com/ostreedev/ostree/issues/1664

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -94,6 +94,16 @@ Boston, MA 02111-1307, USA.
       </varlistentry>
 
       <varlistentry>
+        <term><varname>change-update-summary</varname></term>
+        <listitem><para>Boolean value controlling whether or not to
+        automatically update the summary file after any ref is added,
+        removed, or updated. This covers a superset of the cases covered by
+        commit-update-summary, with the exception of orphan commits which
+        shouldn't affect the summary anyway. Defaults to <literal>false</literal>.
+        </para></listitem>
+      </varlistentry>
+
+      <varlistentry>
         <term><varname>fsync</varname></term>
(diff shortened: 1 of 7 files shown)


mwleeds:
This was proposed upstream in https://github.com/ostreedev/ostree/pull/1681 but that PR doesn't seem to be moving quickly enough so I think we should get it in downstream for now.

mwleeds:
I cherry-picked the commits that made it in upstream, which use a helper function and delay the summary update until the end of transactions.

mwleeds:
I'm going to go ahead and merge this on the basis that:
1. It's already been merged upstream
2. It only adds the option and doesn't enable it by default (see https://github.com/endlessm/eos-image-builder/pull/550)
3. It's difficult for me to test the performance of it without having the option available

PR #135, opened 2018-09-05
Backport min free space API

uajain:


src/libostree/libostree-devel.sym
```diff
@@ -23,6 +23,7 @@ global:
     ostree_mutable_tree_fill_empty_from_dirtree;
     ostree_mutable_tree_new_from_checksum;
     ostree_mutable_tree_check_error;
+    ostree_repo_get_min_free_space_bytes;
```
uajain: Very skeptical about LIBOSTREE_2018 version numbers here. Might cause mismatch up/down stream(until we rebase?). Please advice as necessary

uajain: Note to myself:
Make sure to match and verify with experts of `debian/libostree-1-1.symbols` in debian-branch:
```
ostree_repo_get_min_free_space_bytes@LIBOSTREE_2018.X 2018.X
```

pwithnall: Probably makes sense to create a new section for this symbol, since it’s going to be in the upstream 2018.9 release now that it’s been accepted upstream:
```
LIBOSTREE_2018.9 {
global:
    ostree_repo_get_min_free_space_bytes;
} LIBOSTREE_2018.7;
```

When this patch gets rebased (dropped) when we rebase to the 2018.9 release, the symbol versioning will not change.

That said, I often get symbol versioning wrong, so you’ll have to try my suggestion out and see if it works. The corresponding Debian `libostree-1-1.symbols` line would be something like:
```
ostree_repo_get_min_free_space_bytes@LIBOSTREE_2018.9 2018.6-3endless2
```
inserted in the right place for alphabetical order. The reasoning behind it is that `ostree_repo_get_min_free_space_bytes@LIBOSTREE_2018.9` is the symbol name and version coming from the `.so` file; and `2018.6-3endless2` is the first Debian package version which it appeared in. The idea behind the Debian symbols file is to allow package dependencies to be automatically calculated by looking at symbol dependencies in object files.

uajain: >That said, I often get symbol versioning wrong, so you’ll have to try my suggestion out and see if it works.

Yes, this is very non-trivial to me. Not even sure how to test that your/mine/x solution even "works" or not.

pwithnall: > Not even sure how to test that your/mine/x solution even "works" or not.

Build it!

pwithnall:
bot, retest this please

uajain:
Ah, I have changed the released file `src/libostree/libostree-released.sym` in 
4707c67, therefore the checksum in `tests/test-symbols.sh` is not matching(therefore, failing test). If the change in `src/libostree/libostree-released.sym` looks in the right direction for back-porting, I can go ahead and update the checksum.

pwithnall:
I would be more tempted to leave the `LIBOSTREE_2018.7` section where it is in `src/libostree/libostree-devel.sym`, and add the `LIBOSTREE_2018.9` section there too. Then you don’t need to update the checksum for `libostree-released.sym`.

uajain:
Does the order matter? Because:
```
LIBOSTREE_2018.9 {
 global:
    ... 
} LIBOSTREE_2018.7;

LIBOSTREE_2018.7 {
global:
 ....
} LIBOSTREE_2018.6;
```
Gives me an error;
But the below doesn't
```
LIBOSTREE_2018.7 {
global:
 ...
} LIBOSTREE_2018.6;

LIBOSTREE_2018.9 {
 global:
 ...
} LIBOSTREE_2018.7;
```
Also, is it really OK to have 2 sections?

src/libostree/libostree-devel.sym
pwithnall: No, I think the `LIBOSTREE_2018.9` version tag needs to depend on the `LIBOSTREE_2018.7` tag (which is present in `libostree-devel.sym`), since that tag has been used to control the symbol exports on our most recent OSTree release. The LD version tags have no semantics, and there’s no formal relationship between them and the project version number. 

To put it another way, making `LIBOSTREE_2018.9` depend on `LIBOSTREE_2018.6` would introduce a branch in the LD dependencies which might cause problems with the symbol versioning stored in modules which link against libostree (like flatpak), where they would depend on one branch but not the other.

uajain: I see.
It seems that there are some concrete concepts here and I don't really understand them.

uajain:
I really don't have any alternative. As from my questions and comments, you can figure out I am a naive at releasing/versioning stuff. 

Pushed new changes, seems to build! Can you please eye roll these patches again?
Thanks.

PR #144, opened 2018-09-25
create-usb: Add a --commit option

mwleeds:
Currently on Endless OS, the OSTree ref for the operating system is
something like os/eos/amd64/eos3, so that's what gets passed to `ostree
create-usb` when copying the OS to a USB drive (for offline updates).
However, when eos-updater checks for updates it pulls the metadata for a
candidate commit and in so doing updates that eos3 ref to point to the
partial commit being examined as a potential update rather than the
deployed commit. This causes `ostree create-usb` to fail with an error
like "No such metadata object
7fb045cb2d1f1f3a81bfc157c6128ff443eb56350315b9536bdb56aee0659863.dirtree".

OSTree creates deployment refs that look like "ostree/1/1/0" to maintain
a pointer to the deployed commit, but create-usb can't use these because
it shows up in the summary as just a ref, not a collection-ref.

So this commit adds a --commit option to the create-usb command, so we
can use the appropriate ref but copy the deployed commit rather than a
(potentially partial) update commit.

Closes: #1735
Approved by: cgwalters

### DIFF
--- a/man/ostree-create-usb.xml
+++ b/man/ostree-create-usb.xml
@@ -112,6 +112,15 @@ Boston, MA 02111-1307, USA.
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--commit</option>=COMMIT</term>
+
+                <listitem><para>
+                  Pull COMMIT instead of whatever REF points to. This can only
+                  be used if a single ref is specified.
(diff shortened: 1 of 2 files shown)


mwleeds:
This is a cherry-pick from upstream

PR #147, opened 2018-10-16
Rebase on 2019.1

mwleeds:
This is the current state of upstream master (which hopefully will soon be 2018.9) plus nine downstream commits. The only merge conflict was updating the checksum in `test-symbols.sh`. Since we had so many cherry-picks the delta between this and what we have now on master isn't even that big, about 25 commits.

@pwithnall should the commit "lib/repo-pull: Add more debugging on pull failure" be dropped or sent upstream perhaps?

I still need to do the packaging branch.

https://phabricator.endlessm.com/T24078

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -232,9 +232,8 @@ Boston, MA 02111-1307, USA.
         of finders isn't explicitly specified, either by a consumer of
         libostree API or on the command line. Possible values:
         <literal>config</literal>, <literal>lan</literal>, and
-        <literal>mount</literal> (or any combination thereof). If unset, this
-        defaults to <literal>config;mount;</literal> (since the LAN finder is
-        costly).
+        <literal>mount</literal> (or any combination thereof). If
+        unset, this defaults to <literal>config;lan;mount;</literal>.
         </para></listitem>
       </varlistentry>
     </variablelist>
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2967,7 +2967,7 @@ reload_core_config (OstreeRepo          *self,
 
     /* Fall back to a default set of finders */
     if (configured_finders == NULL)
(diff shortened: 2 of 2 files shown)


mwleeds:
I rebased to the latest upstream master and added a new downstream commit on the end, so we can do LAN updates. An alternative would be to change the repo config on the specific deployments that will do LAN updates (in 2019 likely?) but I don't want to make things hard for us if we eventually want to deploy LAN updates to existing deployments.

mwleeds:
From the CI failure:
```
+ dch -v 2018.6+dev144.8626a04-120 -D master --force-distribution 'Automatic release from git (2018.6+dev144.8626a04-120)'
dch: fatal error at line 1083:
New version specified (2018.6+dev144.8626a04-120) is less than
the current version number (2018.8-3)!  Use -b to force.
```

So I think that's caused by the lack of a packaging branch for this.

pwithnall:
Does this still need reviewing, or is it now out of date? From the [discussion on the issue](https://phabricator.endlessm.com/T24078#639593) it looks like there are some unresolved issues and it’s now out of date wrt Debian?

mwleeds:
This is now rebased on upstream master, which is several commits ahead of 2018.9 but they're all commits we want I think. I also checked to make sure everything on downstream master is either already cherry-picked here or already upstream.

Still left to do here:

- [ ] Check if https://github.com/ostreedev/ostree/commit/a4a49724d6f898fd5e76bd6de49d36f7ed8d237e is going to cause problems for us
- [ ] Check if any changes are needed on the packaging branch
- [ ] Check if the unit tests succeed

pwithnall:
Unless you want review beforehand, I’m going to wait until those remaining to-do items are done before reviewing.

mwleeds:
This is now rebased on upstream master, which is a few commits ahead of 2019.1

mwleeds:
retest this please

PR #149, opened 2019-01-22
Rebase on 2019.1 (packaging)

mwleeds:
This is the packaging side of https://github.com/endlessm/ostree/pull/147. It's a rebase of our `debian-master` branch onto Debian's `debian/2019.1-1` tag. There were a few changes made; see the commit messages for details.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,18 @@
+ostree (2019.1-1endless1) master; urgency=medium
+
+  * Resync Debian packaging changes:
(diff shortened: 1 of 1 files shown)


mwleeds:
> Looks like you already pushed the `Version_2019.1_debian` tag. It's pointed at debian's 2019.1-1 commit, whereas it's supposed to be at the commit that creates the new downstream changelog version. It won't matter in practice, though.

Ah, in that case I did it wrong last time too which was the example I used

PR #150, opened 2019-01-29
grub2: Disable use of grub-mkconfig on Endless

mwleeds:
OSTree recently gained support[1] for recognizing the Debian-style
/boot/grub/grub.cfg path in addition to the existing support for the Red
Hat-style /boot/grub2/grub.cfg. With this, it decides to run
grub-mkconfig in ostree_sysroot_write_deployment_with_options() by
calling ostree_bootloader_write_config(). However grub-mkconfig is
disabled on Endless OS since the grub.cfg is hand-written and doesn't
need updates[2] and therefore returns a non-zero exit status. This
causes operations such as "ostree admin unlock --hotfix" and "ostree
admin upgrade" to fail.

So this commit effectively reverts the Debian grub support, by making
_ostree_bootloader_grub2_query() always return TRUE with out_is_active
set to FALSE, so that grub-mkconfig is not called and the operations
above succeed. In the longer term, Endless should migrate to letting
OSTree manage the grub config and drop this patch[3][4].

https://phabricator.endlessm.com/T25195

[1] https://github.com/ostreedev/ostree/commit/74bdf7e17
[2] https://phabricator.endlessm.com/T19614
[3] https://phabricator.endlessm.com/T14870
[4] https://phabricator.endlessm.com/T18848

### DIFF
--- a/src/libostree/ostree-bootloader-grub2.c
+++ b/src/libostree/ostree-bootloader-grub2.c
@@ -77,6 +77,16 @@ _ostree_bootloader_grub2_query (OstreeBootloader *bootloader,
 {
   OstreeBootloaderGrub2 *self = OSTREE_BOOTLOADER_GRUB2 (bootloader);
 
+  /* FIXME: Endless specific patch: don't let libostree find our
+   * /boot/grub/grub.cfg because we manage that separately and disable
+   * grub-mkconfig. See:
+   * - https://phabricator.endlessm.com/T19614
+   * - https://phabricator.endlessm.com/T18848 */
+  {
+    *out_is_active = FALSE;
+    return TRUE;
+  }
+
   /* Look for the BIOS path first */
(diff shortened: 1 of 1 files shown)


mwleeds:
I haven't managed to solidly test this. I reproduced the bug on a nightly image, unlocked it (by temporarily moving `/boot/grub/grub.cfg`), built and installed this branch, and then used `ostree admin deploy` to try to deploy the rollback deployment. But that of course means on the next boot I had the unpatched version of ostree, so it doesn't make sense that `ostree admin unlock --hotfix` then succeeded.

mwleeds:
Perhaps we should test this after merging to master and before merging to stable

wjt:
If you were able to deploy the rollback deployment, I think that's test enough. The other way to test this would be to unlock a not-latest build (using the workaround you describe), install this package, then perform `eos-updater-ctl upgrade --force` – if it succeeds, all is good.

PR #151, opened 2019-01-31
Revert to version shipped in 3.5.4.

wjt:
We are very close to release, and are scared of finding more regressions.

To revert to an older version of the package, we need a version number that compares newer than the newest version on the branch. We would then re-publish the new version as 2019.1-1endless2.

~ is special-cased in Debian versioning: normally XY > X for all non-empty Y, but X~Y < X for all Y.

**DO NOT MERGE** without further discussion. I am pushing this to test that Jenkins would generate a plausible-looking source package.

https://phabricator.endlessm.com/T25195#674277

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+ostree (2018.6-3endless3) eos; urgency=medium
+
+  * Revert to version shipped in Endless OS 3.5.4.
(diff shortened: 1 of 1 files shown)


mwleeds:
This was accomplished by other means; see https://phabricator.endlessm.com/T25195#674985

PR #153, opened 2019-03-30
tests: Update for glib 2.59.2

mwleeds:
glib 2.59.2 uses a non-breaking space instead of a space to
separate the quantity and unit in g_format_size() so update
our test to handle both a plain space and a non-breaking space.

See https://gitlab.gnome.org/GNOME/glib/issues/1625

Closes: #1818
Approved by: cgwalters


mwleeds:
This is a cherry-pick of an upstream commit that should make the unit tests succeed for us

PR #154, opened 2019-06-26
deploy: Create fake symlinks when real ones fail

ManMower:
When trying to deploy on a vfat partitions, such as the EFI ESP, we
fail to create the loader symlink, and deployment fails.

We'd like to use sd-boot, which only works with the ESP, so we need
a way to deploy without using symlinks.

For a quick and dirty minimally invasive fix, I've made deployment
create a bogus symlink - a text file named loader.sln that contains
the name of the real loader dir.

I've modified systemd-boot to read these bogus symlinks as well.

https://phabricator.endlessm.com/T27040

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -55,6 +55,21 @@
 #define OSTREE_DEPLOYMENT_FINALIZING_ID SD_ID128_MAKE(e8,64,6c,d6,3d,ff,46,25,b7,79,09,a8,e7,a4,09,94)
 #endif
 
+static gboolean
+symlink_fat (const char    *target,
+             int            parent_dfd,
+             const char    *linkpath,
+             GCancellable  *cancellable,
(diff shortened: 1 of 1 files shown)


dsd:
I'm worried that if there's ever a legitimate reason why symlink creation fails on a ext4 partition, we'd fallthrough and create the .sln file which grub doesn't even use. In that case it would be best to fail outright.

Is it feasible to check if we're on FAT, and if so, go straight to the sln code, otherwise continue to always try the symlink & abort if that fails?

Alternatively can you check the return value of symlinkat() and only fall through to the FAT stuff if the return code indicates that it failed because we're on a fs that doesn't support symlinks? Maybe errno gets set to EOPNOTSUPP or something.

ManMower:
Turns out statfs/fstatfs is a thing, I'll see if that reliably determines if we're on a FATFS like the doc indicates it should.

Sadly, symlink failure returns EPERM.

PR #155, opened 2019-07-04
deploy: Create fake symlinks when on FAT filesystems

ManMower:
When trying to deploy on a vfat partitions, such as the EFI ESP, we
fail to create the loader symlink, and deployment fails.

We'd like to use sd-boot, which only works with the ESP, so we need
a way to deploy without using symlinks.

For a quick and dirty minimally invasive fix, I've made deployment
create a bogus symlink - a text file named loader.sln that contains
the name of the real loader dir.

I've modified systemd-boot to read these bogus symlinks as well.

https://phabricator.endlessm.com/T27040

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -31,6 +31,8 @@
 #include <sys/poll.h>
 #include <linux/fs.h>
 #include <err.h>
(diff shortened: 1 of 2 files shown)


dsd:
retest this please

dsd:
retest this please

PR #156, opened 2019-07-08
tests/sizes.js: Fix byte array unpacking

dbnicholson:
Recent GJS changed how byte arrays are unpacked with some assumptions
that they are likely strings. Manually use get_child_value() and
get_byte() to ensure the correct value is parsed when checking the
`ostree.sizes` metadata.

This should really be split into 2 parts since the `deep_unpack()` is upstream but the additional `objType` byte is only downstream. Just looking for review right now.

https://phabricator.endlessm.com/T27213


dbnicholson:
Upstream PR in ostreedev/ostree#1884.

PR #159, opened 2019-09-17
Rebase to ostree 2019.3-2 from debian experimental

andrunko:
Note that I've removed the commits that were only changing debian/changelog to restore old entries as I don't think it is worth rebasing those every time, but please let me know and I will reinstate them if needed.

Here are the details on the changes/patches:
- a711f7f3 Merge in old changelog entries
  - Dropped: only updates changelog
- 4705bb80 Restore ostree-boot files to main ostree package
  - Refreshed - note that this one may not be needed anymore given the upstream bug was resolved - see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=824650
- c4568197 debian: Document Endless packaging changes
  - Refreshed
- 9f2cbd88 Add ostree_repo_get_commit_sizes to symbols file
  - Refreshed
- f1db4a0f Revert "Temporarily disable gjs tests"
  - Refreshed
- 83d108a9 Bring back trivial-httpd subcommand
  - Refreshed
- 3d901064 Update README.Endless with pointer to debian's git
  - Refreshed
- 86fd90a5 Release 2018.4-2endless1
  - Dropped: only updates changelog
- 4154684f Release 2018.6-3endless1
  - Dropped: only updates changelog
- 244bc87e Add backported symbols to debian/libostree-1-1.symbols
  - Dropped: only updates changelog
- 8551a375 Release 2019.1-1endless1
  - Dropped: only updates changelog

New commits:
- 6d845969 Release 2019.3-2endless1

Depends on https://github.com/endlessm/ostree/pull/158.

https://phabricator.endlessm.com/T27607


dbnicholson:
I didn't review this yet, but I just wanted to say that the ostree-boot package change needs careful review to make sure we still ship the bits we want.

andrunko:
bot, retest this please

andrunko:
Ok, this should be fixed now - ended up reverting 9572f7e4f2f9045a0bdc1b621e1e779bbda62e69 and 6262b20f158ae0da1ac93cb21142e82d30a67215 to disable `ostree-boot` (now enabled upstream) and kept our changes to move selected bits of `ostree-boot` to the `ostree` pkg. 

Another possible option could be to ship `ostree-boot`, remove the grub2 files we don't need from it and add the `ostree-boot` package to `eos-meta/os-depends|eos-initramfs-efi-depends`, but I preferred to take the "safe" route as I am not sure this is all we'd need.

Please let me know what you think.

andrunko:
Ah and the current CI failure is due to the missing `Version_*` tags, which will be pushed once merging - see also https://github.com/endlessm/ostree/pull/158#issuecomment-532703091.

andrunko:
bot, retest this please

andrunko:
Pushed the tags and now CI is happy :).

dbnicholson:
> Another possible option could be to ship `ostree-boot`, remove the grub2 files we don't need from it and add the `ostree-boot` package to `eos-meta/os-depends|eos-initramfs-efi-depends`, but I preferred to take the "safe" route as I am not sure this is all we'd need.

This is actually what I'd prefer as it removes a long standing delta from debian. I just wanted to make sure that this was considered, but that's exactly what you've found. We just need to make sure the initramfs parts we want are still shipped but the grub config parts aren't (unless someone does some work to ensure they can be on disk without actually editing our grub.cfg file).

If you wanted to be pedantic, you'd add some `Replaces`/`Breaks` to accommodate the files moving between packages to support package upgrades, but most people don't bother.

andrunko:
> Looks OK to me from a quick pass over it. Two minor comments:
> 
> * In `Revert "ostree-boot: weaken dracut Depends to Recommends"`, the commit message could benefit from a sentence or two saying _why_ it was reverted.
> * Similarly for `Revert "Enable ostree-boot package"`

I think I will take the approach of using ostree-boot as discussed above with @dbnicholson and then we won't need these reverts anymore, but please let me know if you think we should keep these reverts and not use the ostree-boot pkg.

andrunko:
> > Another possible option could be to ship `ostree-boot`, remove the grub2 files we don't need from it and add the `ostree-boot` package to `eos-meta/os-depends|eos-initramfs-efi-depends`, but I preferred to take the "safe" route as I am not sure this is all we'd need.
> 
> This is actually what I'd prefer as it removes a long standing delta from debian. I just wanted to make sure that this was considered, but that's exactly what you've found. We just need to make sure the initramfs parts we want are still shipped but the grub config parts aren't (unless someone does some work to ensure they can be on disk without actually editing our grub.cfg file).
> 
> If you wanted to be pedantic, you'd add some `Replaces`/`Breaks` to accommodate the files moving between packages to support package upgrades, but most people don't bother.

This is already the case for the ostree-boot package from upstream.

andrunko:
Branch updated with the following changes:
- Drop debian/patches - now included in source branch
- Squashed commits updating README.Endless and updated the doc with latest info
- Restored ostree-boot as done by upstream:
  - Removed grub2 files from it and opened https://github.com/endlessm/eos-meta/pull/598 to add ostree-boot to the image

andrunko:
Force pushed to update a commit msg (no actual change).

PR #162, opened 2019-11-04
libostree: Add ostree_async_progress_copy_state()

ptomato:
This allows copying the state from one OstreeAsyncProgress object to
another, atomically, without invoking the callback. This is needed in
libflatpak, in order to chain OstreeAsyncProgress objects so that you
can still receive progress updates when iterating a different
GMainContext than the one that the OstreeAsyncProgress object was
created under.

https://phabricator.endlessm.com/T26965

### DIFF
--- a/src/libostree/libostree-released.sym
+++ b/src/libostree/libostree-released.sym
@@ -551,6 +551,7 @@ LIBOSTREE_2019.2 {
 
 LIBOSTREE_2019.3 {
 global:
+  ostree_async_progress_copy_state;
   ostree_repo_write_archive_to_mtree_from_fd;
   ostree_kernel_args_free;
   ostree_kernel_args_new;
--- a/src/libostree/ostree-async-progress.c
+++ b/src/libostree/ostree-async-progress.c
@@ -425,6 +425,42 @@ ostree_async_progress_set_uint64 (OstreeAsyncProgress       *self,
 }
 
 /**
+ * ostree_async_progress_copy_state:
+ * @self: An #OstreeAsyncProgress to copy from
+ * @dest: An #OstreeAsyncProgress to copy to
+ *
+ * Atomically copies all the state from @self to @dest, without invoking the
+ * callback.
+ * This is used for proxying progress objects across different #GMainContexts.
+ *
+ * Since: 2019.3
+ */
+void
+ostree_async_progress_copy_state (OstreeAsyncProgress *self,
+                                  OstreeAsyncProgress *dest)
+{
+  g_return_if_fail (OSTREE_IS_ASYNC_PROGRESS (self));
+  g_return_if_fail (OSTREE_IS_ASYNC_PROGRESS (dest));
+
+  g_mutex_lock (&self->lock);
+
+  if (self->dead)
+    goto out;
+
+  GLNX_HASH_TABLE_FOREACH_KV (self->values, void *, key, GVariant *, value)
+    {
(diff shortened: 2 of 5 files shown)


src/libostree/libostree-released.sym
```diff
@@ -551,6 +551,7 @@ LIBOSTREE_2019.2 {
 
 LIBOSTREE_2019.3 {
 global:
+  ostree_async_progress_copy_state;
```
pwithnall: Should this not be in a new section of the symbol file?

It also needs to be added to the tests: see the Jenkins failure:
```
Verifying all public symbols are documented:
--- expected-documented.txt	2019-11-04 21:29:16.961623193 +0000
+++ found-documented.txt	2019-11-04 21:29:16.969623259 +0000
@@ -1,4 +1,3 @@
-ostree_async_progress_copy_state
 ostree_async_progress_finish
 ostree_async_progress_get
 ostree_async_progress_get_status
ERROR: tests/test-symbols.sh - too few tests run (expected 3, got 1)
ERROR: tests/test-symbols.sh - exited with status 1
```

src/libostree/ostree-async-progress.c
```diff
@@ -424,6 +424,50 @@ ostree_async_progress_set_uint64 (OstreeAsyncProgress       *self,
...
+/**
+ * ostree_async_progress_copy_state:
+ * @self: An #OstreeAsyncProgress to copy from
+ * @dest: An #OstreeAsyncProgress to copy to
+ *
+ * Atomically copies all the state from @self to @dest, without invoking the
+ * callback.
+ * This is used for proxying progress objects across different GMainContexts.
+ *
+ * Since: 2019.3
```
pwithnall: We’re on 2019.6 now, I think.

src/libostree/ostree-async-progress.c
```diff
@@ -424,6 +424,50 @@ ostree_async_progress_set_uint64 (OstreeAsyncProgress       *self,
...
+}
+
+/**
+ * ostree_async_progress_copy_state:
+ * @self: An #OstreeAsyncProgress to copy from
+ * @dest: An #OstreeAsyncProgress to copy to
+ *
+ * Atomically copies all the state from @self to @dest, without invoking the
+ * callback.
+ * This is used for proxying progress objects across different GMainContexts.
```
pwithnall:
```suggestion
 * This is used for proxying progress objects across different #GMainContexts.
```

src/libostree/libostree-released.sym
ptomato: I think if we add the API without rebasing it would need to go in here, since 2019.3 is the current version on EOS. Upstream it would probably need to go in `libostree-devel.sym` and then get copied into a new section here.

src/libostree/ostree-async-progress.c
ptomato: Not until we rebase it some more, we're currently on 2019.3 ... I am working on the upstream patch and this annotation will have a different version there.

ptomato:
Upstream pull request: https://github.com/ostreedev/ostree/pull/1968

ptomato:
Updated to address upstream review comments, now awaiting further upstream review.

ptomato:
Upstream has not completed a second review, but we decided in the standup today to merge this downstream for 3.7.5 earlier rather than later so that it can receive plenty of testing.

PR #164, opened 2020-02-04
Backport upstream commit sizes APIs

dbnicholson:
This removes the last downstream API we had. It replaces `ostree_repo_get_commit_sizes` with `ostree_commit_get_object_sizes` backported from upstream. This is an ABI break, but the only consumer is eos-updater. It has been prepared for this in https://github.com/endlessm/eos-updater/pull/266 and will need to be rebuilt after this is merged and built.

Note that I added the new functions with the symbol version `LIBOSTREE_2019.7` since that's what they'll have upstream. That way can rebase on 2019.7 without changing the ABI and therefore not have to rebuild eos-updater again.

https://phabricator.endlessm.com/T18171


dbnicholson:
The minor packaging changes are in #165.

dbnicholson:
> I can’t see [any relevant other uses of `size-summary`](https://github.com/search?q=org%3Aendlessm+size-summary&type=Code) in our code, so it looks safe to remove.

I'd never seen anyone use it before and it was actually kind of awkward how it handled commits vs refs relative to other builtins. I'm pretty confident it was just used as a testing interface during the development by Collabora.

PR #166, opened 2020-02-25
deploy: support devicetree directory

starnight:
Add support for a devicetree directory at /usr/lib/modules/$kver/dtb/.
In ARM world a general purpose distribution often suppports multiple
boards with a single operating system. However, OSTree currently only
supports a single device tree, which does not allow to use the same
OSTree on different ARM machines. In this scenario typically the boot
loader selects the effective device tree.

This adds device tree directory support for the new boot artefact
location under /usr/lib/modules. If the file `devicetree` does not
exist, then the folder dtb will be checked. All devicetrees are hashed
into the deployment hash. This makes sure that even a single devicetree
change leads to a new deployment and hence can be rolled back.

The loader configuration has a new key "devicetreepath" which contains
the path where devicetrees are stored. This is also written to the
U-Boot variable "fdt_path". The boot loader is expected to use this path
to load a particular machines device tree from.

Re-diff for https://phabricator.endlessm.com/T29416

Closes: #1900
Signed-off-by: Stefan Agner <stefan.agner@toradex.com>
Signed-off-by: Jian-Hong Pan <jian-hong@endlessm.com>

### DIFF
--- a/src/libostree/ostree-bootloader-uboot.c
+++ b/src/libostree/ostree-bootloader-uboot.c
@@ -144,6 +144,10 @@ create_config_from_boot_loader_entries (OstreeBootloaderUboot     *self,
       if (val)
         g_ptr_array_add (new_lines, g_strdup_printf ("fdt_file%s=%s", index_suffix, val));
 
+      val = ostree_bootconfig_parser_get (config, "devicetreepath");
+      if (val)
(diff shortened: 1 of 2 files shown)


starnight:
Oops!  Build error!

PR #170, opened 2020-06-27
dracut: Create reproducible images

jprvita:
Without reproducible images, a rebuild of the initrd will create a
different image file (due to things like creation time of the files in
the cpio archive) even if the actual contents in it are exactly the
same, adding an unnecessary download during updates.

Adding 'reproducible=yes' avoids this and creates the same image files
for the same content.

https://phabricator.endlessm.com/T30443


mwleeds:
Should probably at least offer this upstream?

jprvita:
> Should probably at least offer this upstream?

Yes, but I want to get this fully tested downstream first

PR #174, opened 2021-01-08
Allow skipping verification binding on pulls

dbnicholson:
2 backports from upstream that add an option to skip verification of commit bindings on pulls. This isn't really needed on master but it's convenient for me for testing things locally that will eventually run on the server. Upstream PRs were ostreedev/ostree#2253 and ostreedev/ostree#2254.


mwleeds (approved):
Looks reasonable

PR #175, opened 2021-01-08
[Infrastructure] Allow skipping verification binding on pulls

dbnicholson:
2 backports from upstream that add an option to skip verification of commit bindings on pulls. This is needed for some work to merge together our current per arch `eos-$arch` repos into one `eos` repo. Our OS repos have many older released commits that lack collection and ref bindings but now live in a repository that has a collection ID. I also found at least one commit that had the wrong ref bindings from the early days when we were getting the bindings in place.

Upstream PRs were ostreedev/ostree#2253 and ostreedev/ostree#2254.

### DIFF
--- a/bash/ostree
+++ b/bash/ostree
@@ -847,6 +847,7 @@ _ostree_pull_local() {
         --gpg-verify-summary
         --require-static-deltas
         --untrusted
+        --disable-verify-bindings
     "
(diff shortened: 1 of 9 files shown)


mwleeds (approved):
Looks reasonable

PR #176, opened 2021-01-29
infra: Fix pull-local with depth

dbnicholson:
This is for the benefit of the repo merging I'm trying to do where I want to pull the entire history of one local repo into another. There were a couple issues running this that were fixed upstream.

Upstream PR at ostreedev/ostree#2267.


wjt:
I'm just going ahead and  merging these since all the patches are all accepted upstream.

PR #177, opened 2021-01-29
Fix pull-local with depth

dbnicholson:
Backports for fixing a local pull with infinite depth. This is the master companion to #176 which is really just for the convenience of local testing.

Upstream PR at ostreedev/ostree#2267.


wjt:
I'm just going ahead and  merging these since all the patches are all accepted upstream.

PR #182, opened 2021-03-22
deploy: Don't try to create symlinks on FAT /boot

ManMower:
Code was introduced to make a symlink from /boot/boot to sysroot,
which breaks for us when /boot is FAT (like on PAYG systems)

Instead of using our mock-symlink code here, just avoid writing
the symlink at all, as it's only for u-boot and syslinux' benefit,
neither of which are used in PAYG.

https://phabricator.endlessm.com/T31593


dbnicholson (approved):
That seems good enough for now.

PR #186, opened 2022-02-04
T32964 rebase 2022.1

dbnicholson:
This is a rebase on top of 2022.1. Here's a list of our commits on master relative to our previous 2020.8:

- [ ] 7610f0cc repo: Delete compat files with commit
- [ ] e7279667 tests: Add tests for Endless compat files

I think it's time to cut the cord on these two. The last time the ostree client would pull these legacy files was in EOS 2.3(!). Even if someone followed the manual EOS 2 to 3 upgrade, the chances that one of the commits that was pulled from that time hasn't been pruned away is very low. I would say the systems this affects is 0.

- [x] 6607b65f lib/repo-pull: Retry pulls without static deltas if they fail

See https://github.com/ostreedev/ostree/pull/1612. I think the last comment from jlebon makes sense but I haven't looked into the feasability. Will file a task to upstream this.

- [x] 1319537e grub2: Disable use of grub-mkconfig on Endless

This is really unfortunate. The upstream way to handle this is to specify the bootloader in the repo config like so:

```
[sysroot]
booloader=none
```

Maybe we should start having the image builder setup repos that way and add another ugly eos-boot-helper script to set it on existing repos so we can someday rely on it. Will file a task to handle that. The other part of reliably saying `bootloader=none` is that we could stop [trimming the ostree-boot package](https://github.com/endlessm/ostree/commit/80cbe1cd90543ba7c02b238cccd76c3f519f4cb5).

- [x] 052f2fd7 deploy: Create fake symlinks when on FAT filesystems
- [x] 09b24be1 deploy: Handle efi blobs

Ugly hacks to make ostree work when `/boot` is a FAT filesystem like when it's the UEFI ESP on PAYG. GNOME OS developer has a nicer solution in https://github.com/ostreedev/ostree/pull/1967, but I don't believe it's ready yet. Will file a downstream task to clean this up.

- [x] 090932dd Skip test-pull-repeated during CI

From debian/patches. We could probably drop this as I've never seen this failure locally or in OBS. From other discussions it sounds like debian's s390x builder has other problems. Leaving it for now.

- [ ] ef2d75cb pull: Allow disabling commit binding verification
- [ ] 1c04dd0d pull: Use GNU coding style
- [ ] 5d130e78 tests: Ensure no dangling commit partials on remote depth pull
- [ ] bec8ceb3 pull: Error on depth pull with missing head commit
- [ ] 514dbada pull: Fix local pull with depth and truncated source history
- [ ] 9533a185 Fix translation of file:// URIs into paths
- [ ] 528e32c0 test-pull-summary-sigs: Set timestamps to serve expected files

Upstream backports.

- [x] 7ce86b9c deploy: Don't try to create symlinks on FAT /boot

Another FAT hack. See above.

- [ ] a8b52f6e repo: Make locking APIs public
- [ ] 19f8684c repo: Require lock type in ostree_repo_lock_pop
- [ ] 62b592e9 build-sys: Bump required GLib to 2.44
- [ ] 4204a7bb repo: Make locking per-OstreeRepo
- [ ] 46e74d3c repo: Make locking precondition failures fatal
- [ ] 7271863f test-concurrency: Lower lock timeout
- [ ] 2c228c8a tests: Add single process repo locking tests

Upstream backports.

https://phabricator.endlessm.com/T32964


dbnicholson:
The post-rebase queue can more clearly be viewed in https://github.com/endlessm/ostree/compare/Version_2022.1...T32964-rebase-2022.1.

dbnicholson:
retest this please

pwithnall (approved):
LGTM.

> From debian/patches. We could probably drop this as I've never seen this failure locally or in OBS. From other discussions it sounds like debian's s390x builder has other problems. Leaving it for now.

I think it makes sense to carry all the `debian/patches` even if, like this one, we’re fairly sure they’re not relevant, as diverging from Debian potentially makes future `debian/patches` harder to apply.

dbnicholson:
retest this please

PR #191, opened 2022-02-07
lib/repo: Add commit version metadata to summary metadata

dbnicholson:
The commit metadata `version` key is well established but getting it for
a remote commit is cumbersome since the commit object needs to be
fetched and loaded. Including it in the summary additional metadata
allows a much more convenient view of what each of the remote refs
represents.

(cherry picked from commit 6fbf759279ca9908e0c7ceb2a3af75b1180fa8a1)

https://phabricator.endlessm.com/T33099

This is only really needed on the server as in #188, but this does include the cosmetic change that the CLI summary view will show `Version` instead of `ostree.commit.version`. The other benefit is to not have server only changes unless they're really needed.


starnight (approved):
LGTM!

PR #196, opened 2022-03-11
Recent ostree backport fixes

dbnicholson:
A couple fixes from upstream that coincide with Debian's [proposed bullseye update](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1006905). The delta commit partial fix is important. The infinite recursion one I haven't seen in practice but the code was clearly wrong before and crash if it was reached.

https://phabricator.endlessm.com/T33209

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -2244,6 +2244,9 @@ process_one_static_delta (OtPullData                 *pull_data,
                                          ref, cancellable, error))
             return FALSE;
 
+          if (!ostree_repo_mark_commit_partial (pull_data->repo, to_revision, TRUE, error))
+            return FALSE;
+
           if (detached_data && !ostree_repo_write_commit_detached_metadata (pull_data->repo,
                                                                             to_revision,
                                                                             detached_data,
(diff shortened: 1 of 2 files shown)


wjt:
Looking at that second commit, don't we also need https://github.com/ostreedev/ostree/pull/2530 ?

dbnicholson:
> Looking at that second commit, don't we also need [ostreedev/ostree#2530](https://github.com/ostreedev/ostree/pull/2530) ?

I discussed this a bit on phabricator but should have included that here.

It doesn't hurt, but we don't technically need it. The 2nd commit fixes `ot_variant_builder_info_free` to not go into an infinite loop when it's called in some conditions. An error in libglnx's `copy_file_range` wrapper when generating deltas is one of those conditions. The fix in libglnx extends the list of errno values that trigger a fallback instead of failing to include `EINVAL`. That helps you get to the fallback on ecryptfs instead of failing. As mentioned [here](https://github.com/ostreedev/ostree/issues/2525#issuecomment-1024089284), they're really 2 separate bugs.

Since we don't support ecryptfs I don't think it's a requirement. However, it's a [very narrow change](https://gitlab.gnome.org/GNOME/libglnx/-/commit/24231a956a4b849087fbf01173cdebb53e1bd60b), so I'm fine if we want to include it, too.

PR #212, opened 2023-12-14
Rebase on 2023.5

dbnicholson:
This rebases on upstream 2023.5. It's not the latest release, but the last few have some hairy changes to do with composefs that I'd rather avoid for now. The remaining commits can be seen [here](https://github.com/endlessm/ostree/compare/Version_2023.5...T35068-rebase-2023.5). Debian branch changes in #213.

- [x] bd4ce64cc lib/repo-pull: Retry pulls without static deltas if they fail

I decided to keep this, but it's possibly not needed anymore. This is a relic of the pre-repo-locking days when it would appear that you would have a full commit but in actuality one of the objects had been pruned away by two processes accessing the repo concurrently. The fix in that scenario is to run `ostree fsck` (or `flatpak repair`), which will mark the commit as partial and it will no longer be considered as a static delta source.

However, we don't have anything that would have guaranteed `ostree fsck` was run. I should also try to revisit the upstream PR and get it in.

- [ ] 344474a91 grub2: Disable use of grub-mkconfig on Endless

Reverted in 6aa9cbc7d since all systems are now guaranteed to specify a non-grub bootloader.

- [x] e178b1c22 deploy: Create fake symlinks when on FAT filesystems
- [x] 60f4a586d deploy: Handle efi blobs
- [x] a8433f120 deploy: Don't try to create symlinks on FAT /boot

Horrible PAYG patches to keep. The 2nd one causes a lot of conflicts that I hope I addressed properly.

- [ ] e20b372a3 Skip test-pull-repeated during CI

Debian has been carrying this for years since their s390x builder occasionally fails there. I've personally run this test locally probably hundreds of times and never had an issue. I doubt we need it, so I'm going to try without it.

- [ ] f05a2607c lib/repo: Add commit version metadata to summary metadata
- [ ] 739fb2459 Fix marking static delta commits as partial
- [ ] a1d9e5632 libotutil: Avoid infinite recursion during error unwinding
- [ ] a379455e9 main: Factor out sysroot loading
- [ ] 97de3c292 finalize-staged: Ensure /boot automount doesn't expire
- [ ] b9bb12757 finalize-staged: Don't listen to `SIGTERM`, just let kernel exit us

Upstream backports.

- [ ] 6aa9cbc7d Revert "grub2: Disable use of grub-mkconfig on Endless"

Companion revert for 344474a91.

- [ ] f9d4eaeac configure: use pkg-config with newer gpgme and gpg-error
[...]

PR #213, opened 2023-12-14
Rebase on Debian 2023.5-1

dbnicholson:
Update the debian branch to 2023.5-1. As detailed below, I decided to drop all of our downstream commits, so there's no further diff! Main branch changes in #212.

- [ ] 3744d4c1c Remove grub2 files from ostree-boot package

As noted on the main branch, all systems are now guaranteed to specify a non-grub bootloader. Therefore, the grub integration will never be used and we don't need to care about whether the files are on disk or not.

- [ ] e3b8d73f2 Revert "Build using FUSE 3"
- [ ] f0bbc8b90 Revert "ostree Breaks flatpak-builder (<< 1.2.1-2~)"

Temporary changes to handle flatpak-builder being built against fuse2 that were reverted below.

- [ ] 4cdf13d96 debian: Document Endless packaging changes

After dropping the grub change, there's no longer any significant delta from Debian. Therefore, there's no reason to keep a `debian/README.Endless` file.

- [ ] 9a0080dde Remove debian/patches

I decided not to bother with this since the debian 1.0 format source packages jenkins-obs creates don't attempt to apply these patches. I decided not to apply the 2 patches to the main branch since they're both for handling test failures on s390x. We're not going to have that issue.

- [ ] 9ce27514e Release 2022.1-3endless1

Gratuitous changelog entry.

- [ ] 104a4984a Revert "Revert "Build using FUSE 3""
- [ ] 5002a34b8 Revert "Revert "ostree Breaks flatpak-builder (<< 1.2.1-2~)""
- [ ] 8d5a07ef0 d/README.Endless: Remove FUSE 2 vs 3 delta

Companion reverts for the above flatpak-builder/fuse2 temporary changes.

- [ ] 30c7d8c92 ostree-boot.install: Add ostree-finalize-staged-hold.service unit

In Debian since 2022.6-1.

- [ ] c6269669a Build with libcurl http backend instead of libsoup2.4

Upstream backport.

https://phabricator.endlessm.com/T35068

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,9 +1,10 @@
-ostree (2023.5-1) UNRELEASED; urgency=medium
+ostree (2023.5-1) unstable; urgency=medium
 
   * Team upload
   * New upstream release
+  * deiban/libostree-1-1.symbols: Add new symbols
 
- -- Jeremy Bícha <jbicha@ubuntu.com>  Wed, 16 Aug 2023 09:47:18 -0400
+ -- Jeremy Bícha <jbicha@ubuntu.com>  Wed, 16 Aug 2023 10:01:24 -0400
 
 ostree (2023.3-2) unstable; urgency=medium
 


starnight (approved):

starnight:
Should the original master branch been archived before force push?

wjt:
I don't think there's any particular reason to. We could recover it from the previous pull request if needed

dbnicholson:
retest this please

dbnicholson:
I'm not sure how useful it is, but I decided to just tag current debian-master as `endless/2022.1-debian` for archiving.

PR #214, opened 2024-09-07
switchroot: Stop making /sysroot mount private

dbnicholson:
Back in 2b8d586c5, /sysroot was changed to be a private mount so that submounts of /var do not propagate back to the stateroot /var. That's laudible, but it makes /sysroot different than every other shared mount in the root namespace. In particular, it means that submounts of /sysroot do not propagate into separate mount namespaces.

Rather than make /sysroot private, make /var a slave+shared mount so that it receives mount events from /sysroot but not vice versa. That achieves the same effect of preventing /var submount events from propagating back to /sysroot while allowing /sysroot mount events to propagate forward like every other system mount. See mount_namespaces(7)[1] and the linux shared subtrees[2] documentation for details on slave+shared mount propagation.

When /var is mounted in the initramfs, this is accomplished with mount(2) syscalls. When /var is mounted after switching to the real root, the mount propagation flags are applied as options in the generated var.mount unit. This depends on a mount(8) feature that has been present since util-linux 2.23. That's available in RHEL 7 and every non-EOL Debian and Ubuntu release. Applying the propagation from var.mount fixes a small race, too. Previously, if a /var submount was added before /sysroot was made private, it would have propagated back into /sysroot. That was possible since ostree-remount.service orders itself after var.mount but not before any /var submounts.

1. https://man7.org/linux/man-pages/man7/mount_namespaces.7.html
2. https://docs.kernel.org/filesystems/sharedsubtree.html

Fixes: #2086

(cherry picked from commit 2973ec591008be94d74d08807079d242b914dcd2 without the test since the kola tests aren't run for Endless)

https://phabricator.endlessm.com/T35640

This is a backport from upstream PR ostreedev/ostree#3292. I left out the kola test since we don't run those.


starnight (approved):
Thanks!

